PR #131, opened 2015-10-02
Fix a TypeError when using a copy of dictionary fields.

wedaly:
Previously, when using a copy of a dictionary of fields,
a TypeError would be raised for invalid schemas, since
the copy of `UNDEFINED` would not be the same object.

This commit changes the comparison to `isinstance` to ensure
that copies of `UNDEFINED` are correctly recognized as undefined.

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -1,3 +1,4 @@
+import copy
 from nose.tools import assert_equal
 
 import voluptuous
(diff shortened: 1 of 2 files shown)


wedaly:
Using a copy of fields is useful for defining several schemas that differ only by a couple of fields.  I want to be able to define a dictionary with the shared fields, copy it, and update it with new fields before creating a schema.

It would be fantastic to get this reviewed, merged, and released to PyPi so I can clean up the code I'm working on.

Thanks for maintaining this awesome tool!

PR #136, opened 2015-12-08
Schema extension

hoxu:
As discussed in issue #135.


hoxu:
Oh well, this was the only portable way to list the resulting schema (keys) I found:

    +>>> sorted(list(person_with_age.schema.keys()))
    +['age', 'name']

*cringe*

PR #177, opened 2016-06-08
provide basic humanize-error wrappers for #140

thatneat:
The humanization wrapper includes the invalid value in the message and also combines all of the messages for a MultipleInvalid.
Landing this would satisfy #140

### DIFF
--- a/voluptuous/humanize.py
+++ b/voluptuous/humanize.py
@@ -1,0 +1,39 @@
+from voluptuous import Invalid, MultipleInvalid
+from voluptuous.error import Error
+
+
+MAX_VALIDATION_ERROR_ITEM_LENGTH = 500
+
+
+def _nested_getitem(data, path):
+    for item_index in path:
+        try:
+            data = data[item_index]
+        except (KeyError, IndexError):
+            # The index is not present in the dictionary, list or other indexable
+            return None
+    return data
+
+
+def humanize_error(data, validation_error, max_sub_error_length=MAX_VALIDATION_ERROR_ITEM_LENGTH):
+    """ Provide a more helpful + complete validation error message than that provided automatically
+    Invalid and MultipleInvalid do not include the offending value in error messages,
+    and MultipleInvalid.__str__ only provides the first error.
+    """
+    if isinstance(validation_error, MultipleInvalid):
+        return '\n'.join(sorted(
+            humanize_error(data, sub_error, max_sub_error_length)
+            for sub_error in validation_error.errors
+        ))
+    else:
+        offending_item_summary = repr(_nested_getitem(data, validation_error.path))
+        if len(offending_item_summary) > max_sub_error_length:
+            offending_item_summary = offending_item_summary[:max_sub_error_length - 3] + '...'
+        return '%s. Got %s' % (validation_error, offending_item_summary)
+
+
+def validate_with_humanized_errors(data, schema, max_sub_error_length=MAX_VALIDATION_ERROR_ITEM_LENGTH):
+    try:
+        return schema(data)
+    except (Invalid, MultipleInvalid) as e:
+        raise Error(humanize_error(data, e, max_sub_error_length))
--- a/voluptuous/tests/tests.py
+++ b/voluptuous/tests/tests.py
@@ -6,6 +6,7 @@ from voluptuous import (
     Url, MultipleInvalid, LiteralInvalid, NotIn, Match, Email,
     Replace, Range, Coerce, All, Any, Length, FqdnUrl, ALLOW_EXTRA, PREVENT_EXTRA
 )
+from voluptuous.humanize import humanize_error
 
 
 def test_required():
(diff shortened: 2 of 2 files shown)


voluptuous/humanize.py
```diff
@@ -0,0 +1,39 @@
...
+    for item_name in path:
+        try:
+            data = data[item_name]
+        except KeyError:
+            return None
+    return data
+
+
+def humanize_error(data, validation_error, max_sub_error_length=MAX_VALIDATION_ERROR_ITEM_LENGTH):
+    ''' Provide a more helpful + complete validation error message than that provided automatically
```
tusharmakkar08: Use triple double quoted string in docstring. Refer PEP 257

voluptuous/humanize.py
```diff
@@ -0,0 +1,39 @@
+from voluptuous import Invalid, MultipleInvalid
+from voluptuous.error import Error
+
+
+MAX_VALIDATION_ERROR_ITEM_LENGTH = 500
+
+
+def _nested_getitem(data, path):
+    for item_name in path:
+        try:
```
tusharmakkar08: Use `if item_name in data` instead of exception handling.

voluptuous/humanize.py
```diff
@@ -0,0 +1,39 @@
+from voluptuous import Invalid, MultipleInvalid
+from voluptuous.error import Error
+
+
+MAX_VALIDATION_ERROR_ITEM_LENGTH = 500
+
+
+def _nested_getitem(data, path):
```
tusharmakkar08: You can use something like: `reduce(operator.getitem, path, data)`

voluptuous/humanize.py
```diff
@@ -0,0 +1,39 @@
...
+        return '\n'.join(sorted(
+            humanize_error(data, sub_error, max_sub_error_length)
+            for sub_error in validation_error.errors
+        ))
+    else:
+        offending_item = _nested_getitem(data, validation_error.path)
+        offending_item_summary = repr(offending_item)
+        if len(offending_item_summary) > max_sub_error_length:
+            offending_item_summary = offending_item_summary[:max_sub_error_length - 3] + '...'
+        return '%s. got %s' % (validation_error, offending_item_summary)
```
tusharmakkar08: Change to `Got`

voluptuous/humanize.py
```diff
@@ -0,0 +1,39 @@
...
+    and MultipleInvalid.__str__ only provides the first error.
+    """
+    if isinstance(validation_error, MultipleInvalid):
+        return '\n'.join(sorted(
+            humanize_error(data, sub_error, max_sub_error_length)
+            for sub_error in validation_error.errors
+        ))
+    else:
+        offending_item = _nested_getitem(data, validation_error.path)
+        offending_item_summary = repr(offending_item)
```
tusharmakkar08: Use message chaining.

tusharmakkar08:
Hey @thatneat : Thanks for the PR. I have reviewed the PR. There are few comments. Please fix them and squash the commits.

voluptuous/humanize.py
thatneat: Would you prefer this or the functional `reduce` method mentioned above? Actually, it occurs to me that missing-value handling might not be required here, and if it is, we probably need to handle IndexError as well.

voluptuous/humanize.py
thatneat: Not sure what you mean here. Are you looking for a conditional expression like:
'''python
return recursive_thing() if isinstance(validation_error, MultipleInvalid) else the_other_thing()
'''

voluptuous/humanize.py
thatneat: I'd probably lean against the `reduce` option even though it's nice and concise, for python3 compatibility (though I know it's not officially supported)

voluptuous/humanize.py
thatneat: or are you looking for a `partial` + `map` instead of a generator expression?
I didn't think there was a need to deal with nested MultipleInvalids

tusharmakkar08: I mean to just have :  `offending_item_summary = repr(_nested_getitem(data, validation_error.path))` instead of 2 separate variable initialization

voluptuous/humanize.py
tusharmakkar08: Ok. Let's avoid `reduce` due to python3 compatibility issue. However it would be good if you can use conditionals instead of exceptionals.

thatneat: not using EAFP here will make the code more complex since we're dealing with lists/tuples/other indexables in addition to dicts, but I'll give it a shot.

thatneat: here's what I ended up with; it's kind of the posterchild of why EAFP is sometimes more straightforward and thus safer in Python:
```python
def _nested_getitem(data, path):
    for item_index in path:
        # Data could be a dict or indexable sequence (list/tuple)
        item_present = (
            isinstance(data, dict) and item_index in data
            or item_index > (len(data) - 1)
        )
        if item_present:
            data = data[item_index]
        else:
            return None
    return data
```

I can go with that if you prefer (it does have the advantage of being more explicit about lists - maybe I should add a test case for that anyway) but I think the `try: except` is more straightforward because you don't have to reverse-engineer what's going to actually happen when you try to index.

tusharmakkar08: Yeah this indeed looks complicated. In `except` clause add 'IndexError' along with comment explaining the implication.

thatneat:
@tusharmakkar08 thanks for reviewing! I think everything's good to go now.

tusharmakkar08:
@thatneat 🍰 👍

thatneat:
@tusharmakkar08 / @alecthomas any idea when this can get into a release (probably 0.9.0 due to the added functionality)?

PR #205, opened 2016-09-16
Fixes #204: Making Path validators resilient to None.

### LINKED ISSUE #204: IsFile and IsDir and other path validators breaks on None
```py
from voluptuous import IsFile
s = IsFile()
s('.bashrc') # ok
s('jkfhsdh') # throws: FileInvalid: not a file
s(None) # throws TypeError: coercing to Unicode: need string or buffer, NoneType found
```
I think in case of the value is None, the voluptuous has to raise FileInvalid exception

tusharmakkar08:
Fixes https://github.com/alecthomas/voluptuous/issues/204

### DIFF
--- a/voluptuous/validators.py
+++ b/voluptuous/validators.py
@@ -416,8 +416,13 @@ def IsFile(v):
     True
     >>> with raises(FileInvalid, 'not a file'):
     ...   IsFile()("random_filename_goes_here.py")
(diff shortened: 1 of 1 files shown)


tusharmakkar08:
@alecthomas : Build has passed. Let's get this merged.

PR #207, opened 2016-09-21
Feature

nareshnootoo:
Varify the maximum number of digits that are present in the number(Precision),
    and the maximum number of decimal places(Scale)

    :raises Invalid: If the value does not match the provided Precision and Scale.

    >>> s = Schema(Number(precision=6, scale=2))

### DIFF
--- a/voluptuous/tests/tests.py
+++ b/voluptuous/tests/tests.py
@@ -5,7 +5,7 @@ from voluptuous import (
     Schema, Required, Extra, Invalid, In, Remove, Literal,
     Url, MultipleInvalid, LiteralInvalid, NotIn, Match, Email,
     Replace, Range, Coerce, All, Any, Length, FqdnUrl, ALLOW_EXTRA, PREVENT_EXTRA,
-    validate, ExactSequence, Equal, Unordered
+    validate, ExactSequence, Equal, Unordered, Number
 )
 from voluptuous.humanize import humanize_error
 
@@ -557,3 +557,101 @@ def test_schema_decorator_return_only_unmatch():
         return "hello"
 
     assert_raises(Invalid, fn, 1)
+
+
+def test_number_validation_with_string():
+    """ test with Number with string"""
+    schema = Schema({"number" : Number(precision=6, scale=2)})
+    try:
+        schema({"number": 'teststr'})
+    except MultipleInvalid as e:
+        assert_equal(str(e),
+                     "Value must be a number enclosed with string for dictionary value @ data['number']")
+    else:
+        assert False, "Did not raise Invalid for String"
+
+
+def test_number_validation_with_invalid_precision_invalid_scale():
+    """ test with Number with invalid precision and scale"""
+    schema = Schema({"number" : Number(precision=6, scale=2)})
+    try:
+        schema({"number": '123456.712'})
+    except MultipleInvalid as e:
+        assert_equal(str(e),
+                     "Precision must be equal to 6, and Scale must be equal to 2 for dictionary value @ data['number']")
+    else:
+        assert False, "Did not raise Invalid for String"
+
+
+def test_number_validation_with_valid_precision_scale():
+    """ test with Number with valid precision and scale"""
+    schema = Schema({"number" : Number(precision=6, scale=2, yield_decimal=True)})
+    out_ = schema({"number": '1234.00'})
+    assert_equal(float(out_.get("number")), 1234.00)
+
+
+def test_number_when_precision_scale_none():
+    """ test with Number with no precision and scale"""
+    schema = Schema({"number" : Number(yield_decimal=True)})
+    out_ = schema({"number": '12345678901234'})
+    assert_equal(out_.get("number"), 12345678901234)
+
+
+def test_number_when_precision_none_n_valid_scale_case1():
+    """ test with Number with no precision and valid scale case 1"""
+    schema = Schema({"number" : Number(scale=2, yield_decimal=True)})
+    out_ = schema({"number": '123456789.34'})
+    assert_equal(float(out_.get("number")), 123456789.34)
+
+
+def test_number_when_precision_none_n_valid_scale_case2():
+    """ test with Number with no precision and valid scale case 2 with zero in decimal part"""
+    schema = Schema({"number" : Number(scale=2, yield_decimal=True)})
+    out_ = schema({"number": '123456789012.00'})
+    assert_equal(float(out_.get("number")), 123456789012.00)
+
+
+def test_number_when_precision_none_n_invalid_scale():
+    """ test with Number with no precision and invalid scale"""
+    schema = Schema({"number" : Number(scale=2, yield_decimal=True)})
+    try:
+        schema({"number": '12345678901.234'})
+    except MultipleInvalid as e:
+        assert_equal(str(e),
+                     "Scale must be equal to 2 for dictionary value @ data['number']")
+    else:
+        assert False, "Did not raise Invalid for String"
+
+
+def test_number_when_valid_precision_n_scale_none():
+    """ test with Number with no precision and valid scale"""
+    schema = Schema({"number" : Number(precision=14, yield_decimal=True)})
+    out_ = schema({"number": '1234567.8901234'})
+    assert_equal(float(out_.get("number")), 1234567.8901234)
+
+
+def test_number_when_invalid_precision_n_scale_none():
+    """ test with Number with no precision and invalid scale"""
+    schema = Schema({"number" : Number(precision=14, yield_decimal=True)})
+    try:
+        schema({"number": '12345674.8901234'})
+    except MultipleInvalid as e:
+        assert_equal(str(e),
+                     "Precision must be equal to 14 for dictionary value @ data['number']")
+    else:
+        assert False, "Did not raise Invalid for String"
+
+
+def test_number_validation_with_valid_precision_scale_yield_decimal_none():
+    """ test with Number with valid precision, scale and no yield_decimal"""
+    schema = Schema({"number" : Number(precision=6, scale=2)})
+    out_ = schema({"number": '1234.00'})
+    assert_equal(out_.get("number"), '1234.00')
+
+
+def test_number_validation_with_valid_precision_scale_yield_decimal_false():
+    """ test with Number with valid precision, scale and no yield_decimal"""
+    schema = Schema({"number" : Number(precision=6, scale=2, yield_decimal=False)})
+    out_ = schema({"number": '1234.00'})
+    assert_equal(out_.get("number"), '1234.00')
+
--- a/voluptuous/tests/tests.py
+++ b/voluptuous/tests/tests.py
@@ -583,35 +583,35 @@ def test_number_validation_with_invalid_precision_invalid_scale():
         assert False, "Did not raise Invalid for String"
 
 
-def test_number_validation_with_valid_precision_scale():
+def test_number_validation_with_valid_precision_scale_yield_decimal_true():
     """ test with Number with valid precision and scale"""
     schema = Schema({"number" : Number(precision=6, scale=2, yield_decimal=True)})
     out_ = schema({"number": '1234.00'})
     assert_equal(float(out_.get("number")), 1234.00)
 
 
-def test_number_when_precision_scale_none():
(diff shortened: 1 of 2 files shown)


voluptuous/validators.py
```diff
@@ -794,3 +794,44 @@ def __call__(self, v):
...
+    """Limit a value to a Number.
+
+    :raises Invalid: If the value does not match the provided Precision and Scale.
+
+    >>> s = Schema(Number(precision=6, scale=2))
+
+    resource : http://stackoverflow.com/questions/3018758/determine-precision-and-scale-of-particular-number-in-python
+    """
+
+    def __init__(self, precision=None, scale=None, msg=None, max_digits=14):
```
tusharmakkar08: make `max_digit` constant

voluptuous/tests/tests.py
```diff
@@ -557,3 +557,31 @@ def fn(arg):
         return "hello"
 
     assert_raises(Invalid, fn, 1)
+
```
tusharmakkar08: Add following test cases: 

- [ ]  scale is `None`
- [ ]  precision is `None`
- [ ]  Both are `None`

voluptuous/validators.py
```diff
@@ -794,3 +794,44 @@ def __call__(self, v):
...
+        if magnitude >= self.max_digits:
+            return (magnitude, 0)
+        frac_part = abs(v) - int_part
+        multiplier = 10 ** (self.max_digits - magnitude)
+        frac_digits = multiplier + int(multiplier * frac_part + 0.5)
+        while frac_digits % 10 == 0:
+            frac_digits /= 10
+        scale = int(math.log10(frac_digits))
+
+        if magnitude + scale > self.precision or scale > self.scale :
```
tusharmakkar08: Make 2 separate `Invalid` conditions

voluptuous/validators.py
```diff
@@ -794,3 +794,44 @@ def __call__(self, v):
...
+            raise Invalid(self.msg or "Precision must be smaller than %s, and Scale must be smaller than %s"%(self.precision, self.scale))
```
tusharmakkar08: This will break when `self.precision is None` or `self.scale is None`

nareshnootoo:
@tusharmakkar08 I have made some changes as per your suggestion please review

voluptuous/tests/tests.py
```diff
@@ -557,3 +557,61 @@ def fn(arg):
...
+        schema({"number": 123456.712})
+    except MultipleInvalid as e:
+        assert_equal(str(e),
+                     "Precision must be smaller than 6, and Scale must be smaller than 2 for dictionary value @ data['number']")
+    else:
+        assert False, "Did not raise Invalid for String"
+
+def test_number_validation_with_valid_number():
+    """ test with Number with valid precision and scale"""
+    schema = Schema({"number" : Number(precision=6, scale=2)})
```
tusharmakkar08: Shouldn't this yield error?

voluptuous/validators.py
```diff
@@ -794,3 +794,60 @@ def __call__(self, v):
 
     def __repr__(self):
         return 'Unordered([{}])'.format(", ".join(repr(v) for v in self.validators))
+
+
+class Number(object):
+    """
+    Varify the maximum number of digits that are present in the number(Precision),
```
tusharmakkar08: *Verify

voluptuous/validators.py
```diff
@@ -794,3 +794,60 @@ def __call__(self, v):
...
+
+            if self.scale is not None and scale > self.scale :
+                raise Invalid(self.msg or "Scale must be smaller or equal to %s"%self.scale)
+
+        return v
+
+    def __repr__(self):
+        return ('Number(precision=%s, scale=%s, msg=%s)' % (self.precision, self.scale, self.msg))
+
+    def get_precision_scale(self, number):
```
tusharmakkar08: private function

voluptuous/validators.py
```diff
@@ -794,3 +794,60 @@ def __call__(self, v):
...
+        """
+        resource : http://stackoverflow.com/questions/3018758/determine-precision-and-scale-of-particular-number-in-python
+        :param number:
+        :return: tuple(precision, scale)
+        """
+        max_digits = 14
```
tusharmakkar08: Better make it default argument in the class.

voluptuous/tests/tests.py
nareshnootoo: In which scenario?

tusharmakkar08: I mean this scenario should give error and the correct number should be : 1234.00

nareshnootoo: I think we shouldn't be so hard, as 1234 is a valid number, however we can return it with decimal places, what do you think?

tusharmakkar08: But the point of having a separate `Number` validator is to tackle these kind of issues only. If we want 1234, directly to be accepted then instead of using `Number` we can use `int` which is already supported by voluptuous.

nareshnootoo: Make sense :)

nareshnootoo: But in python if I declare a variable as `a=1.000` it always give me `1.0`
same in case of PHP it return me as `1`
also in general 1.00 == 1

tusharmakkar08: You can try some different implementation. Something on the lines of using  [Decimal](https://docs.python.org/2/library/decimal.html) and string conversions.

tusharmakkar08:
Hey @nareshnootoo,

Thanks

nareshnootoo:
Hi @tusharmakkar08 I have implemented the Decimal and string conversions. Please review

voluptuous/tests/tests.py
```diff
@@ -557,3 +557,68 @@ def fn(arg):
...
+    try:
+        schema({"number": 'teststr'})
+    except MultipleInvalid as e:
+        assert_equal(str(e),
+                     "Value must be a number enclosed with string for dictionary value @ data['number']")
+    else:
+        assert False, "Did not raise Invalid for String"
+
+def test_number_validation_with_invalid_precision_scale():
```
tusharmakkar08: Add a test case with only precision and no scale.

voluptuous/validators.py
```diff
@@ -794,3 +793,57 @@ def __call__(self, v):
...
+
+class Number(object):
+    """
+    Verify the maximum number of digits that are present in the number(Precision),
+    and the decimal places(Scale)
+
+    :raises Invalid: If the value does not match the provided Precision and Scale.
+
+    >>> schema = Schema({"number" : Number(precision=6, scale=2)})
+    >>> out_ = schema({"number": '1234.01'})
```
tusharmakkar08: Add an assert for doctests here as well.

voluptuous/validators.py
```diff
@@ -794,3 +793,57 @@ def __call__(self, v):
...
+
+            if self.scale is not None and scale != self.scale :
+                raise Invalid(self.msg or "Scale must be equal to %s"%self.scale)
+
+        return number
+
+    def __repr__(self):
+        return ('Number(precision=%s, scale=%s, msg=%s)' % (self.precision, self.scale, self.msg))
+
+    def __get_precision_scale(self, number):
```
tusharmakkar08: Let's make `__get_precision_scale` internal instead of `_get_precision_scale`. Refer http://stackoverflow.com/a/6930223/1799475

voluptuous/validators.py
```diff
@@ -794,3 +793,57 @@ def __call__(self, v):
...
+        """
+        :param number:
+        :return: tuple(precision, scale, decimal_number)
+        """
+        from decimal import Decimal
```
tusharmakkar08: Move this import up.

voluptuous/validators.py
```diff
@@ -794,3 +793,57 @@ def __call__(self, v):
...
+
+        try:
+            decimal_num = Decimal(number)
+        except:
```
tusharmakkar08: Please avoid having generic Exception catching.

voluptuous/validators.py
```diff
@@ -806,7 +806,7 @@ class Number(object):
 
     >>> schema = Schema(Number(precision=6, scale=2))
     >>> schema('1234.01')
-    1234.01
+    Decimal('1234.01')
```
tusharmakkar08: Assert here

nareshnootoo:
Hi @tusharmakkar08 I have fixed the following:
1) Moved Decimal import up
2) Added InvalidOperation exception instead of generic
3) Made __get_precision_scale to internal as _get_precision_scale
4) Added assert for doctests
5) Added case with only invalid precision and no scale

voluptuous/validators.py
```diff
@@ -794,3 +794,57 @@ def __call__(self, v):
...
+class Number(object):
+    """
+    Verify the maximum number of digits that are present in the number(Precision),
+    and the decimal places(Scale)
+
+    :raises Invalid: If the value does not match the provided Precision and Scale.
+
+    >>> schema = Schema(Number(precision=6, scale=2))
+    >>> schema('1234.01')
+    Decimal('1234.01')
```
tusharmakkar08: Add assert here.

voluptuous/tests/tests.py
```diff
@@ -557,3 +557,78 @@ def fn(arg):
...
+    assert_equal(float(out_.get("number")), 123456789.34)
+
+def test_number_when_precision_none_n_valid_scale_2():
+    """ test with Number with no precision and valid scale case 2 with zero in decimal part"""
+    schema = Schema({"number" : Number(scale=2)})
+    out_ = schema({"number": '123456789012.00'})
+    assert_equal(float(out_.get("number")), 123456789012.00)
+
+def test_number_when_precision_none_n_invalid_scale():
+    """ test with Number with no precision and invalid scale"""
```
tusharmakkar08: It's difficult to keep track of all the function names in tests since few of them are like `test_number_when_precision_none_n_invalid_scale` while others are like `test_number_when_scale_none_n_valid_precision`. Please follow a consistent convention over here, preferably add precision first and then scale.

voluptuous/validators.py
tusharmakkar08: Can we do something about Decimal output ? (Maybe have string output). @alecthomas What do you think about this?

nareshnootoo: Yes we can have string output, but user can use processed Decimal.

tusharmakkar08: @nareshnootoo  Let's add a flag which tells what to yield, where default being the string since we are giving input as string only.

voluptuous/tests/tests.py
```diff
@@ -557,3 +557,101 @@ def fn(arg):
...
+    try:
+        schema({"number": '12345674.8901234'})
+    except MultipleInvalid as e:
+        assert_equal(str(e),
+                     "Precision must be equal to 14 for dictionary value @ data['number']")
+    else:
+        assert False, "Did not raise Invalid for String"
+
+
+def test_number_validation_with_valid_precision_scale_yield_decimal_none():
```
tusharmakkar08: yield_decimal_false

tusharmakkar08:
Hey @nareshnootoo 

The commits haven't been squashed (Also fix the comment in my last review). After doing rebase, you might need to do `git push -f` since if you will pull, it will pull all the old commits as well. 

Thanks

nareshnootoo:
@tusharmakkar08 I have squashed the commits.

tusharmakkar08:
@nareshnootoo : Please check the earlier comment.

voluptuous/tests/tests.py
nareshnootoo: @tusharmakkar08 there is already a test case named test_number_validation_with_valid_precision_scale_yield_decimal_false

tusharmakkar08: I mean add for yield_decimal_true since this test is same as test at Line 652.

tusharmakkar08:
Hey @nareshnootoo 

Please squash the changes and get this merged into master 🍰 

Thanks

nareshnootoo:
Hi @tusharmakkar08 I have squashed the commits.

tusharmakkar08:
Merged 🍰

nareshnootoo:
Thanks :)

PR #209, opened 2016-09-22
Adding eq for schemas

tusharmakkar08:
Fixes https://github.com/alecthomas/voluptuous/issues/208


tusharmakkar08:
I think there is some issue in here. Will fix it and reopen the PR.

PR #224, opened 2016-09-28
Prevent the 'u' prefix on keys

### LINKED ISSUE #218: Remove the u prefix of unicode strings on error output
I use unicode everywhere within my project, and this affects the output of Voluptuous' exceptions.

`required key not provided @ data[u'site_user_id']`

Is it possible to remove the unicode prefix?

jmwri:
Fixes #218

### DIFF
--- a/voluptuous/tests/tests.py
+++ b/voluptuous/tests/tests.py
@@ -1,13 +1,14 @@
 import copy
-from nose.tools import assert_equal, assert_raises
+from nose.tools import assert_equal, assert_raises, assert_true
 
 from voluptuous import (
     Schema, Required, Extra, Invalid, In, Remove, Literal,
     Url, MultipleInvalid, LiteralInvalid, NotIn, Match, Email,
     Replace, Range, Coerce, All, Any, Length, FqdnUrl, ALLOW_EXTRA, PREVENT_EXTRA,
-    validate, ExactSequence, Equal, Unordered
+    validate, ExactSequence, Equal, Unordered, Number
 )
 from voluptuous.humanize import humanize_error
+from voluptuous.util import to_utf8_py2, u
 
 
 def test_exact_sequence():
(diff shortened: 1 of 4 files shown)


tusharmakkar08:
Hey @jmwri 

Thanks

jmwri:
@tusharmakkar08 Sorry about that, had some issues with 3.2. I've added six to install_requires.

voluptuous/encoding.py
```diff
@@ -0,0 +1,5 @@
+def to_utf8_py2(data):
+    import sys
```
tusharmakkar08: Move this import to top

setup.py
```diff
@@ -50,5 +50,6 @@
     ],
     install_requires=[
         'setuptools >= 0.6b1',
+        'six >= 1.10'
```
tusharmakkar08: Add it in requirements.txt

voluptuous/encoding.py
```diff
@@ -0,0 +1,5 @@
+def to_utf8_py2(data):
```
jmwri: I'm having some trouble with this. When I move the function I'm getting lots of doctest errors.

requirements.txt
```diff
@@ -0,0 +1,2 @@
+nose==1.3.7
```
tusharmakkar08: You will have to update `setup.py` also for adding `requirements.txt` to install_requires.

voluptuous/encoding.py
tusharmakkar08: What errors are you getting ? Is it due to circular dependency?

voluptuous/tests/tests.py
```diff
@@ -8,6 +8,8 @@
     validate, ExactSequence, Equal, Unordered
 )
 from voluptuous.humanize import humanize_error
+from voluptuous.encoding import to_utf8_py2
+import six
```
tusharmakkar08: Can we get this done without importing `six`? Because Ideally we don't want voluptuous to have external dependencies. Check [this](https://docs.python.org/2/howto/unicode.html#the-unicode-type)

jmwri: I've had to make separate tests for py2 & 3 due to checking for unicode or str objects.

requirements.txt
jmwri: No problem

tusharmakkar08:
@jmwri : There are few conflicts. Please resolve the conflicts and update the PR.

requirements.txt
tusharmakkar08: We can now remove `requirements.txt` since this is not a dependency on voluptuous but a dependency on tests.

setup.py
```diff
@@ -22,6 +22,13 @@
 
 description = long_description.splitlines()[0].strip()
 
+try:
```
tusharmakkar08: Please revert it back to original setup.py

setup.py
```diff
@@ -22,7 +22,6 @@
 
 description = long_description.splitlines()[0].strip()
 
-
```
tusharmakkar08: Please revert this change as well

voluptuous/util.py
```diff
@@ -148,3 +150,17 @@ def __str__(self):
...
         return repr(self.lit)
+
+
+def to_utf8_py2(data):
+    if sys.version_info < (3,) and isinstance(data, unicode):
+        return data.encode('utf-8')
+    return data
+
+
+if sys.version_info < (3,):
```
tusharmakkar08: Why not doing something like: 

```
def u(x):
      if sys.version_info < (3, ):
           return unicode(x)
      else:
          return x
```

tusharmakkar08:
@jmwri : Merged 🍰

jmwri:
Yay!

PR #231, opened 2016-10-09
Validate namedtuples as tuples

### LINKED ISSUE #230: NamedTuples do not validate as tuples or named tuples
```
from collections import namedtuple

NT = namedtuple('NT', ['a', 'b'])

nt = NT(1, 2)
t = (1, 2)

Schema((int, int))(nt)    # => Invalid
Schema((int, int))(t)     # => Valid
Schema(NT(int, int))(nt)  # => Invalid
Schema(NT(int, int))(t)   # => Valid
````

Given that `NT(1, 2) == (1, 2) => True` I'd expect all of the above cases to return as valid.

dangitall:
namedtuples can't be initialized with a tuple. Detect that datatype and
initialize it with *args notation.

Resolves #230

### DIFF
--- a/voluptuous/schema_builder.py
+++ b/voluptuous/schema_builder.py
@@ -106,6 +106,10 @@ ALLOW_EXTRA = 1  # extra keys not in schema will be included in output
 REMOVE_EXTRA = 2  # extra keys not in schema will be excluded from output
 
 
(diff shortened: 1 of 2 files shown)


tusharmakkar08:
Merged 🍰

Issue #254, opened 2016-12-07
validating an OrderedDict returns an unordered dict


SimplyKnownAsG:
It would be nice if the returned dictionary from validation were the same type as the original.

For example, if I validate an OrderedDict of (x,y) pairs, the order may be important and converting to a traditional dict doesn't maintain the order. I suppose it is also possible that one could request the validator be ordered as well and I'm not sure what that would mean.

Example:

```python
import numpy
import collections
from voluptuous import Schema, Number

schema = Schema({Number(): Number()}) # x, y pairs (for interpolation or something)
data = collections.OrderedDict(zip(numpy.linspace(5, 100, 11), numpy.random.random(11)))
out = schema(data)
assert isinstance(out, collections.OrderedDict),\
    'Collection is no longer ordered, order:{}'.format(', '.join(str(k) for k in out.keys()))
```
---

This may be related to #40, but I don't quite understand what the issue was there.

PR #255, opened 2016-12-07
Added support for OrderedDict validation, including as the return type

SimplyKnownAsG:
This addresses #254, which I created earlier. The fix is pretty simple, sorry for the stupid formatting stuff, I can get rid of those changes as  you desire.

One thing I thought was going to be an isuses was ensuring that `data` is a `dict`; however, that is already done in the `validate_dict` closure (right?).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ MANIFEST
 .idea
 venv
 docs/_build
+.vscode/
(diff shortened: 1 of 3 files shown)


SimplyKnownAsG:
@alecthomas, the CI failed because Python2.6 does not have a `collections.OrderedDict`. Is it acceptable to disable the test if `collections.OrderedDict` does not exist?

Issue #259, opened 2017-01-17
Validation of a dictionary runs in O(n^2)


odedfos:
Currently all keys in the dictionary are matched against all keys in the schema. If we have a large set of keys then the validation takes very long.

Here is a test that simulates the problem. A single validation should take under 1 ms but currently takes 12 ms.

```
def test_validation_performance():
    """
    Naive validation of a dictionary with many keys with a schema that has many keys
     This test verifies a single validation runs in under 1 msec
    :return:
    """
    number_if_keys_in_schema = 100
    iterations = 1000
    threshold_ms = 1

    schema_dict = {str(i): str for i in range(number_if_keys_in_schema)}

    schema = Schema(schema_dict)

    data = {str(i): str(i) for i in range(number_if_keys_in_schema)}

    start = time.time()

    for i in xrange(iterations):
        schema(data)

    total_elapsed = time.time() - start
    avg_per_round = round(total_elapsed * 1000 / iterations, 3)
    print("Total: %s secs, avg: %s ms" % (round(total_elapsed, 3), avg_per_round))

    assert avg_per_round < threshold_ms, "Avg validation %s > %s ms" % (avg_per_round, threshold_ms)
```

odedfos:
Thanks!
When is a pypi release expected?

tusharmakkar08:
Hey @odedfos 

We need to update `CHANGELOG.md` for the corresponding changes. Once it's done, we'll release it. Add your changes in the `CHANGELOG.md` and create a corresponding PR. We are planning to release new version in the early February. 

Thanks.

odedfos:
@tusharmakkar08, Any update on the package release?

PR #260, opened 2017-01-17
#259 Fix performance issue of large dict validation

odedfos:
A fix for performance issue [#259](https://github.com/alecthomas/voluptuous/issues/259)


voluptuous/tests/tests.py
```diff
@@ -788,10 +787,47 @@ def test_ordered_dict():
...
+    data_extra_keys = {}
+
+    counter = [0]
+
+    class CounterMarker(Marker):
+        def __call__(self, *args, **kwargs):
+            counter[0] += 1
+            return super(CounterMarker, self).__call__(*args, **kwargs)
+
+    for i in range(num_of_keys):
```
tusharmakkar08: Let's use `xrange` instead

odedfos: xrange is not compatible with python >= 3
For simplicity of the tests I think it might be better to stick with 'range'. What do you think?

tusharmakkar08: Cool. Looks fine.

PR #285, opened 2017-04-30
Allow all schemas in Maybe

balloob:
The `Maybe` validator is only limited to types now. With this change it will be backwards compatible but also allow any type of schema to be passed in.

My only concern is that i don't know what to do with the msg parameter and so for my initial version I have removed it. We could make `__call__` be something like this:

```python
try:
    return self.schema(v)
except Invalid:
    if self.msg is None:
        raise
    raise Invalid(self.msg)
```

### DIFF
--- a/voluptuous/tests/tests.py
+++ b/voluptuous/tests/tests.py
@@ -552,8 +552,12 @@ def test_maybe():
     s = Schema(Maybe(int))
     assert s(1) == 1
     assert s(None) is None
-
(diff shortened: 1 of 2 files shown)


balloob:
Done

balloob:
The drop in coverage that Coveralls reports is not related to the `Maybe` validator: https://coveralls.io/builds/11310155/source?filename=voluptuous%2Fvalidators.py#L474

PR #294, opened 2017-06-26
Drop support for end-of-life Python versions

jonafato:
Per the [dev guide](https://docs.python.org/devguide/#status-of-python-branches), the following versions have been end of lifed:

- 2.6: 2013-10-29
- 3.1: 2012-04-11
- 3.2: 2016-02-20

Removing support for EOL versions should make future maintenance easier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-- '2.6'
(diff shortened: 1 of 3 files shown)


jonafato:
Conflicts resolved.

PR #314, opened 2017-12-03
Adding SomeOf validator

guyarad:
The SomeOf validator can verify that some of validators are met, or optionally at most number of validators.


voluptuous/validators.py
```diff
@@ -933,3 +934,49 @@ def _get_precision_scale(self, number):
             raise Invalid(self.msg or 'Value must be a number enclosed with string')
 
         return (len(decimal_num.as_tuple().digits), -(decimal_num.as_tuple().exponent), decimal_num)
+
+
+class SomeOf(object):
+    """Value must pass at least some validations, determined by the given parameter
+
+    The output of each validator is passed as input to the next.
```
guyarad: description will be improved in the next update of this PR (I'm sure there will be so I'm waiting with pushing it 😄 )

guyarad:
Is this repository not active anymore?

tusharmakkar08:
Hey @guyarad 

I was busy over the weekends. Will review it and get back to you. Before my review, please fix the failing tests. 

Thanks.

guyarad:
@tusharmakkar08 no rush - just wanted to know how active we are 😊 

the failing tests aren't mine. they seem to exist after the last commit to master.

voluptuous/validators.py
tusharmakkar08: Haha.. Let's not rely on it.

voluptuous/validators.py
```diff
@@ -933,3 +934,49 @@ def _get_precision_scale(self, number):
...
+    >>> validate(6.6)
+    6.6
+    >>> validate(3)
+    3
+    >>> with raises(MultipleInvalid, 'value must be at most 5, not a valid value'):
+    ...     validate(6.2)
+    """
+
+    def __init__(self, min_valid, validators, max_valid=None, **kwargs):
+        self.min_valid = min_valid or 0
```
tusharmakkar08: Why `min_valid` is not a default keyword argument while `max_valid` is?

tusharmakkar08:
Hey @guyarad 

Thanks for pointing out the reason behind failing tests. Have created an issue for the same:  https://github.com/alecthomas/voluptuous/issues/315 . 

I have glanced over the PR. Have a couple of comments. Let's resolve them and get this merged. 

Thanks.

voluptuous/validators.py
guyarad: Originally I only needed "minimum valid schemas" and just added "max valid" for completeness. So for me, min_valid was always needed. We can put them both as optional, and assert that either of them has a value (otherwise, what's the point of this validator). 

Sounds good?

voluptuous/validators.py
guyarad: Will push it asap

.python-version
```diff
@@ -0,0 +1,5 @@
+2.7.14
+3.6.3
+3.3.6
+3.4.7
+3.5.2
```
guyarad: this file was accidentally added because I was working with `pyenv` which proved to be very useful for testing locally with multiple python versions. 

I was going to remove it, but then thought it would be a good to have such convention as repository users, to use pyenv. your call.

voluptuous/tests/tests.py
```diff
@@ -968,3 +968,36 @@ def test_description():
...
+            validator('a')
+
+        with raises(MultipleInvalid, 'no uppercase letters, no lowercase letters'):
+            validator('wqs2!#s111')
+
+        with raises(MultipleInvalid, 'no lowercase letters, no symbols'):
+            validator('3A34SDEF5')
+
+
+def test_SomeOf_max_validation():
```
tusharmakkar08: Let's add a test for checking the assert statement too using `AssertionError`

.python-version
tusharmakkar08: @alecthomas What do you think about adding this to the package?

voluptuous/tests/tests.py
guyarad: Done

.python-version
tusharmakkar08: Let's remove this for now and create an issue where we will continue this discussion.

Issue #315, opened 2017-12-06
Travis Failing Test Cases in Master


tusharmakkar08:
Hey

Merging of PR: https://github.com/alecthomas/voluptuous/pull/311 has resulted in tests failure in the master branch. 

@dtao: Please look into this.

dtao:
Yikes! I will take a look.

dtao:
Hmm, so now that I'm looking into this I'm noticing that the issue seems to be that the `Schema.__eq__` method just tests for *string* equality of two objects' `schema` attributes. Since ordering of keys is not deterministic for `dict`s, I would argue this is actually a bug (i.e. if `a == Schema({'foo': 1, 'bar': 2})` and `b == Schema({'foo': 1, 'bar': 2})`, then `a == b` is non-deterministic). In which case the fix here is probably to fix the `__eq__` method to be deterministic for dict-based schemas.

@alecthomas Does that sound right to you? I am happy to create a PR to fix the issue.

PR #316, opened 2017-12-06
make Schema.__eq__ deterministic

dtao:
This adds a test to highlight the issue raised in #315 and proposes a solution: instead of comparing `str(self.schema)` with `str(other)`, just compare `schema` attributes directly.

### DIFF
--- a/voluptuous/schema_builder.py
+++ b/voluptuous/schema_builder.py
@@ -246,6 +246,9 @@ class Schema(object):
             return False
         return other.schema == self.schema
 
+    def __ne__(self, other):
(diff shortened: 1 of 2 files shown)


dtao:
Update: after adding some more tests for this, I noticed the the `__ne__` method was not defined; so I implemented that as well. (Now there are quite a few tests that fail before these changes!)

alecthomas:
@tusharmakkar08 you should be able to Squash and merge the branch through GitHub:

image

tusharmakkar08:
@alecthomas Didn't know it existed 😅

PR #330, opened 2018-02-16
Don't stop to first error in a list of objects

tuukkamustonen:
List of primitives iterates whole list and records every error, but list of objects doesn't.

The change is quite radical - I don't really understand what the removed block does, but removing it doesn't fail any tests. I think it's only in effect when validating a list of objects. But why was it originally added - why stop at first error and not validate all elements in a list?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [Unreleased]
+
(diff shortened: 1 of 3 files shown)


tuukkamustonen:
Seems like doctest failed, I'll check it later on when I more time.

PR #348, opened 2018-07-05
Include path in AnyInvalid errors

### LINKED ISSUE #347: Undocumented breaking change between 0.10.5 and 0.11.1 regarding error handling
Following the migration from 0.10.5 to 0.11.1, there has been a change in the way error are handled that is not documented on the changelog.

As one can see below, the `path` part of the error (for `q2`) is missing when a error message is specified in `0.11.1` while it used to be filled in `0.10.5`.

```
[root@localhost vagrant]# cat vol.py
#!/bin/python                          
                                                                                                                                                                                              
import voluptuous as v   

_schema = v.Schema({
    v.Optional('q'): int,
    v.Required('q2'): v.Any(int, msg='toto')
})

try:
    _schema({'q': 'str', 'q2': 'tata'})
except v.MultipleInvalid as exc:
     print(str([d.__dict__ for d in exc.errors]))

[root@localhost vagrant]# python -c "import voluptuous; print(voluptuous.__version__)"
0.10.5

[root@localhost vagrant]# python vol.py
[{'path': ['q'], 'error_type': 'dictionary value', 'error_message': 'expected int'}, {'path': ['q2'], 'error_type': 'dictionary value', 'error_message': 'toto'}]

[root@localhost vagrant]# python -c "import voluptuous; print(voluptuous.__version__)"                                                                                                        
0.11.1

[root@localhost vagrant]# python vol.py
[{'path': ['q'], 'error_type': 'dictionary value', 'error_message': 'expected int'}, {'path': [], 'error_type': 'dictionary value', 'error_message': 'toto'}]
```

jd:
Fixes #347

### DIFF
--- a/voluptuous/tests/tests.py
+++ b/voluptuous/tests/tests.py
@@ -1085,6 +1085,34 @@ def test_self_validation():
     schema({"follow": {"follow": {"number": 123456}}})
 
 
+def test_any_error_has_path():
+    """https://github.com/alecthomas/voluptuous/issues/347"""
+    s = Schema({
+        Optional('q'): int,
+        Required('q2'): Any(int, msg='toto')
+    })
+    try:
+        s({'q': 'str', 'q2': 'tata'})
+    except MultipleInvalid as exc:
+        assert exc.errors == [
+            {'path': ['q'],
+             'error_type': 'dictionary value',
+             'error_message': 'expected int'},
+            {'path': ['q2'],
+             'error_type': 'dictionary value',
+             'error_message': 'toto'},
+        ] or [
+            {'path': ['q2'],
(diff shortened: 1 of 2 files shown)


jd:
Invalid, needs to be fixed.

jd:
Pull-request updated, HEAD is now 3c48ef9e1e7bcb0211e58d35d2fc46bda0feca22

Spredzy:
@jd @alecthomas Since this fixes a breaking change between `0.10.5` and `0.11.1`, could a new tag be published ? This way user update don't have to face this regression.

Spredzy:
@alecthomas up

PR #372, opened 2018-12-14
Ignore .pytest_cache in .gitignore

svisser:
Allow anyone using py.test to have the pytest cache ignored. I noticed this when working on https://github.com/alecthomas/voluptuous/pull/371.


svisser:
Cheers, will try to find the right balance between small changes and changes that require further input 👍

PR #411, opened 2019-11-07
Ensure Maybe propagates error information

monopolis:
Due to Any evaluating all options, and raising the exception
from the first encountered error, Maybe would discard the
expected error raised by the provided validator.

This commit changes the order of the validators in Maybe so
that the first evaluated error (and thus returned) is that of
the provided validator.

### DIFF
--- a/voluptuous/tests/tests.py
+++ b/voluptuous/tests/tests.py
@@ -514,7 +514,7 @@ def test_repr():
     )
     assert_equal(repr(coerce_), "Coerce(int, msg='moo')")
     assert_equal(repr(all_), "All('10', Coerce(int, msg=None), msg='all msg')")
-    assert_equal(repr(maybe_int), "Any(None, %s, msg=None)" % str(int))
+    assert_equal(repr(maybe_int), "Any(%s, None, msg=None)" % str(int))
 
 
 def test_list_validation_messages():
@@ -1526,5 +1526,21 @@ def test_any_with_discriminant():
         })
     except MultipleInvalid as e:
         assert_equal(str(e), 'expected bool for dictionary value @ data[\'implementation\'][\'c-value\']')
+    else:
+        assert False, "Did not raise correct Invalid"
+
+def test_maybe_returns_subvalidator_error():
+    schema = Schema(Maybe(Range(1, 2)))
+
+    # The following should be valid
+    schema(None)
+    schema(1)
+    schema(2)
+
+    try:
+        # Should trigger a MultipleInvalid exception
+        schema(3)
+    except MultipleInvalid as e:
+        assert_equal(str(e), "value must be at most 2")
     else:
         assert False, "Did not raise correct Invalid"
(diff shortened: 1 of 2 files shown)


svisser:
Hmm, it's interesting because the test case now fails in Python 3.x due to:

```python
  File "/home/travis/build/alecthomas/voluptuous/voluptuous/validators.py", line 589, in __call__
    if self.min is not None and not v >= self.min:
TypeError: '>=' not supported between instances of 'NoneType' and 'int'
```

This previously didn't happen as it checked for `None` first.

monopolis:
Please see pull request #414 for a potential fix for the incomparable values. It should solve this issue at the very least.

My guess is that there probably are other validators that should check for TypeError's as well  -  e.g. Length comes to mind.

voluptuous/tests/tests.py
```diff
@@ -1437,9 +1437,25 @@ def test_any_with_discriminant():
...
+def test_maybe_returns_subvalidator_error():
+    schema = Schema(Maybe(Range(1, 2)))
+
+    # The following should be valid
+    schema(None)
+    schema(1)
+    schema(2)
+
+    try:
+        # Should trigger a UrlInvalid exception
```
spacegaier, line 1456: Comment incorrectly says `UrlInvalid`, but I think should say `MultipleInvalid`.

spacegaier:
> 
> 
> Please see pull request #414 for a potential fix for the incomparable values. It should solve this issue at the very least.
> 
> My guess is that there probably are other validators that should check for TypeError's as well - e.g. Length comes to mind.

@svisser The mentioned PR #414 has by now been merged and indeed fixes this issue for me.
@monopolis I added additional TypeErrors to Clamp and Length, so that is covered now.

voluptuous/tests/tests.py
monopolis: Hi!

Thank you for taking the time to review this!

I have addressed the comment, and rebased the commit on upstream/master.

voluptuous/tests/tests.py
```diff
@@ -1528,3 +1528,19 @@ def test_any_with_discriminant():
         assert_equal(str(e), 'expected bool for dictionary value @ data[\'implementation\'][\'c-value\']')
     else:
         assert False, "Did not raise correct Invalid"
+
+def test_maybe_returns_subvalidator_error():
```
spacegaier, line 1532: Can you move that up to where the other Maybe test are so that they grouped together?

voluptuous/tests/tests.py
```diff
@@ -1528,3 +1528,19 @@ def test_any_with_discriminant():
...
+    # The following should be valid
+    schema(None)
+    schema(1)
+    schema(2)
+
+    try:
+        # Should trigger a MultipleInvalid exception
+        schema(3)
+    except MultipleInvalid as e:
+        assert_equal(str(e), "value must be at most 2")
```
spacegaier, line 1544:
```suggestion
        assert_equal(str(e), "value must be None or in the range from 1 to 2")
```

voluptuous/tests/tests.py
spacegaier: I will merge as-is and then update the grouping via a separate change.

Issue #419, opened 2020-03-30
`Required` flag not set properly after #379 fix


EarQuack:
Following test (this will be a new test, if accepted) fails:

```python
def test_new_required_test():
    schema = Schema({
        'mykey': All(int, Range(1, 20)),
    }, required=True)
    assert_true(schema.required)
```

Suspect problem was introduced in #379 

Will propose a solution and issue pull request to fix

EarQuack:
@alecthomas Please assign me #419 and I will propose solution and passing test.  Best regards!

tusharmakkar08:
Assigned @EarQuack

PR #420, opened 2020-03-31
Fixed issue with 'required' not being set properly and added test

EarQuack:
A fix for issue #419 - schema required flag isn't set properly in all cases


tusharmakkar08:
LGTM

PR #422, opened 2020-06-26
Switch to pytest and include Python 3.10

pgajdos:
I am not completely sure about tox.ini etc.

### DIFF
--- a/voluptuous/tests/tests.py
+++ b/voluptuous/tests/tests.py
@@ -19,7 +19,7 @@ def test_new_required_test():
     schema = Schema({
         'my_key': All(int, Range(1, 20)),
     }, required=True)
-    assert_true(schema.required)
+    assert schema.required
 
 
 def test_exact_sequence():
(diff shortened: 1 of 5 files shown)


spacegaier:
What would the benefit of that switch? I haven't looked myself into the differences between `nosetest` and `pytest`, but I assume you have. Could you provide your reasoning for advocating that change in `voluptuous`?

pgajdos:
nose is deprecated
https://nose.readthedocs.io/en/latest/index.html

voluptuous/tests/tests.py
```diff
@@ -31,18 +30,16 @@ def test_exact_sequence():
...
     schema = Schema({Required('q'): 1})
-    # Can't use nose's raises (because we need to access the raised
-    # exception, nor assert_raises which fails with Python 2.6.9.
     try:
         schema({})
     except Invalid as e:
-        assert_equal(str(e), "required key not provided @ data['q']")
+        assert str(e) == "required key not provided @ data['q']"
     else:
         assert False, "Did not raise Invalid"
```
freundTech, line 44: You could use `pytest.raises` here and for all other exceptions:

```
with pytest.raises(Invalid) as e:
    schema({})
assert str(e.value) == "required key not provided @ data['q']"

spacegaier:
Superseded by https://github.com/alecthomas/voluptuous/pull/464

PR #423, opened 2020-06-30
Fix serializer KeyError while asserting Schema

anshukch:
After DRF upgrade to 3.6.4 (ymmv), serializer data uses ReturnDict which threw
error while asserting the API response data. To fix this, the hack was required
to proxy it to OrderedDict instead. Doctests fail now. But it is mostly because
the expected isn't OrderedDict format. Yet the values are correct.

### DIFF
--- a/voluptuous/schema_builder.py
+++ b/voluptuous/schema_builder.py
@@ -589,7 +589,8 @@ class Schema(object):
 
             if errors:
                 raise er.MultipleInvalid(errors)
-
+            from collections import OrderedDict
+            data = OrderedDict(data)
             out = data.__class__()
             return base_validate(path, iteritems(data), out)
 


anshukch:
Okay. I accidentally created this PR against your repo 😄 . But I anyway was gonna discuss this issue with you.

Basically, after DRF upgrade to 3.6.4 `Schema` `assert` are failing as serializer data is `ReturnDict` instance and throws `KeyError: 'serializer'`

In order to Bypass that I did a hack after forking it, which you may ignore obviously. Also because the doctests syntactically fail but are functionally unaffected.

I can close the PR. But I leave it open if you want to carry forward this conversation. Looking forward to hear from you.

spacegaier:
I am not sure if I can follow what exactly is the issue here. Could you provide an example I can have a look at?

PR #425, opened 2020-07-24
Improve error message for In() and NotIn()

vmaillol-altair:
Improve Error raised by In validator


spacegaier:
@vmaillol-altair If that is added then it probably should be also for `class NotIn`.

Issue #451, opened 2022-03-18
Do not use sorted()


Avorthoren:
Version 12.1 (commit https://github.com/alecthomas/voluptuous/pull/436/commits/111e1c25c6bcfbbd0da9fdb4fcc1370ffa111169) adds call of `sorted` function. This leads to application crash when schema container contains not comparable pair of elements. Example:
```
>>> import voluptuous as vlps
>>> s = vlps.Schema(vlps.In((str, int, float)))
>>> d = s(42)
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 3, in <module>
  File ".../lib/python3.9/site-packages/voluptuous/schema_builder.py", line 272, in __call__
    return self._compiled([], data)
  File ".../lib/python3.9/site-packages/voluptuous/schema_builder.py", line 817, in validate_callable
    return schema(data)
  File ".../lib/python3.9/site-packages/voluptuous/validators.py", line 741, in __call__
    'value must be one of {}'.format(sorted(self.container)))
TypeError: '<' not supported between instances of 'type' and 'type'
```

This function should not be used, as it limits the possibilities of use without giving anything in return.

PR #452, opened 2022-03-19
Call of sorted() removed from In and NotIn validators, tests adjusted.

Avorthoren:
https://github.com/alecthomas/voluptuous/issues/451 resolved


spacegaier:
I created an alternative PR https://github.com/alecthomas/voluptuous/pull/463 which fixes your reported issue while still being able to keep the sorting as we have it currently.

PR #457, opened 2022-04-01
Enable github actions

epenet:
Enable github actions to run tox sessions

I have not enabled flake8 or py27 or py310 as these currently fail.


epenet:
Sample run: https://github.com/epenet/voluptuous/actions/runs/2075801822

PR #458, opened 2022-04-01
Adjust documentation for running tests

epenet:
`nose-cover3` relies on `use_2to3` for which support was removed in `setuptools 58`.
See https://setuptools.pypa.io/en/latest/history.html#v58-0-0

It may be necessary to first downgrade `setuptools` to a previous version when installing `nose-cover3`:
```console
foo@bar:~$ pip install "setuptools<58"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -715,9 +715,15 @@ s({'password':'123', 'password_again': 1337})
 
 ## Running tests
 
-Voluptuous is using nosetests:
+Voluptuous is using `nosetests`:
 
-    $ nosetests
+    $ pip install nose nose-cover3 "coverage>=3.0"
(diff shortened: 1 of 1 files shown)


spacegaier:
`nose-cover3` would not be required if only `nosetests` is run without coverage analysis, correct?

So perhaps, we should mention both approaches, the one without coverage but then also not enforcing the `setuptools` version, and below the more complete tests with coverage (so basically the `tox` definition) and the additional info from this PR?

spacegaier:
With the switch to `pytest` (https://github.com/alecthomas/voluptuous/pull/464) this PR is now obsolete. I created a similar one https://github.com/alecthomas/voluptuous/pull/465 to provide some more info reg. the `pytest` runs.

PR #459, opened 2022-04-01
Enable flake8 in github actions

epenet:
Follow-up to #457

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,7 +3,8 @@ envlist = flake8,py27,py36,py37,py38,py39
 
 [flake8]
 ; E501: line too long (X > 79 characters)
-ignore = E501
+; W504 line break after binary operator
+ignore = E501,W504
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -3,12 +3,14 @@ envlist = flake8,py27,py36,py37,py38,py39
 
 [flake8]
 ; E501: line too long (X > 79 characters)
-ignore = E501
+; W504 line break after binary operator
+ignore = E501,W504
```
epenet, line 7: We need to choose between ignoring W503 or W504
I would prefer to ignore W503 and enable W504, but current code is formatted with line break after so...

spacegaier: I will merge this PR as-is with W504 being ignored, and push a second PR to change that, since I already locally switched to ignoring W503 while testing the flake8 setup.

PR #460, opened 2022-04-01
Enable python 2.7 in github actions

epenet:
Follow-up to #457
Needs #454


epenet:
This is now part of #454

PR #461, opened 2022-04-01
Enable python 3.10 in github actions

epenet:
Follow-up to #457
Needs #422

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,39 @@
+name: Tests
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  tests:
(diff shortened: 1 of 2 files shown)


epenet:
Or maybe we just drop coverage from the py310 session in tox...

epenet:
Of maybe we do need to switch to pytest if the issue is inside nosetest

spacegaier:
> Of maybe we do need to switch to pytest if the issue is inside nosetest

I think switching makes sense, also given that `nose` is now deprecated. And if we have to switch `pytest` makes probably more sense than `nose2`.

spacegaier:
Superseded by https://github.com/alecthomas/voluptuous/pull/464

PR #463, opened 2022-04-02
Ensure `sorted` in `In` works with container of data types

spacegaier:
Alternative to https://github.com/alecthomas/voluptuous/pull/452

I would prefer to keep the sorting, since that IMHO improves the error message we return, especially when utilized in user facing use cases.

This PR nonetheless ensures we properly handle containers with data types.

### DIFF
--- a/voluptuous/tests/tests.py
+++ b/voluptuous/tests/tests.py
@@ -81,7 +81,7 @@ def test_in():
         assert False, "Did not raise InInvalid"
 
 
-def test_in_class():
+def test_in_data_type_container():
     """Verify that In works with container of data types."""
     schema = Schema({"type": In((int, str, float))})
     schema({"type": float})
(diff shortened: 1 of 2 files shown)


Avorthoren:
Do you really want to sort, for example, `int`s by their `str` representantion?
```
>>> print(sorted([1, 11, 2], key=str))
[1, 11, 2]
```

beastd:
```py
def __call__(self, v):
        try:
            check = v not in self.container
        except TypeError:
            check = True
        if check:
            if self.msg:
                raise InInvalid(self.msg)

            try:
                allowed = sorted(self.container)
            except:
                allowed = sorted(self.container, key=str)

            raise InInvalid('value must be one of {}'.format(allowed))
        return v
```

Maybe something like this would be acceptable?

Not strictly related to the issue at hand: Maybe containers with lots of items should neither be sorted nor rendered in the help message. One could e.g. just say "value must be one of 123456 allowed values" instead.

PR #500, opened 2023-12-13
refactor(tests): fix few tests, use pytest.raises, extend raises helper

antoni-szych-rtbhouse:
Hi, I thought it would be nice to modernize tests a bit:

- use `with pytest.raises()` or `with raises()` in multiple places
- parametrize a few tests
- make `raises()` manager throw AssertionError if expected exception was not raised inside the context
- fix 1 or 2 asserts in tests that were incorrectly working due to the above
- remove python 2 or <3.2 compatibility code (*note: this extends beyond tests*)

I have some doubts regarding existing tests if they really test what they say they do, especially `test_self_any` and `test_self_all`. However it should be checked by someone with more experience (I haven't used `Self`-related validation yet myself)

### DIFF
--- a/voluptuous/schema_builder.py
+++ b/voluptuous/schema_builder.py
@@ -13,25 +13,6 @@ from collections.abc import Generator
 import typing
 from voluptuous.error import Error
 
-if sys.version_info >= (3,):
-    long = int
-    unicode = str
-    basestring = str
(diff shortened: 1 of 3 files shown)


alecthomas (approved):
Good idea!

PR #503, opened 2023-12-14
Add linters configuration, reformat whole code

antoni-szych-rtbhouse:
Hi,

As black and isort became the standard tools to enforce uniform style within the whole codebase, I've came to a conclusion that these may be helpful also here, as the style currently varies in different parts of the codebase.

I've added:
- black
- isort
- moved the pytest config to pyproject.toml

I've reformated the code according to black and isort configuration.

Topics that are very open for discussion:
- line lengths. Black gives default of 88 characters per line. ~Given that nearly all displays currentlyt have 16:9 / 16:10 aspect ratio instead of 4:3, 120 characters per line also seems reasonable (which I've set in the configuration in pyproject.toml)~ **Default of 88 chars is used**. However this setting is entirely dependent on the personal preference and can be easily changed to any other value. [more reading](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length)
- string normalization. Black by default converts all strings that do not contain double quotes `"` to double-quoted ones instead of singe-quoted `'`. However, some people have strong opinion regarding quoting style and allow double-quotes only in docstrings. ~I've left the default normalization on, but performed the normalization in separate commit. If needed, that commit can be easily skipped and the configuration can be changed to include `skip-string-normalization`~. [more reading](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings)
String normalization was dropped from this PR.

### DIFF
--- a/voluptuous/__init__.py
+++ b/voluptuous/__init__.py
@@ -1,9 +1,10 @@
 # flake8: noqa
 
 from voluptuous.schema_builder import *
-from voluptuous.validators import *
 from voluptuous.util import *
-from voluptuous.error import *
+from voluptuous.validators import *
+
+from voluptuous.error import * # isort: skip
 
 __version__ = '0.14.1'
 __author__ = 'alecthomas'
--- a/voluptuous/__init__.py
+++ b/voluptuous/__init__.py
@@ -1,10 +1,12 @@
 # flake8: noqa
-
+# fmt: off
 from voluptuous.schema_builder import *
 from voluptuous.util import *
 from voluptuous.validators import *
 
 from voluptuous.error import * # isort: skip
+
+# fmt: on
 
 __version__ = '0.14.1'
 __author__ = 'alecthomas'
--- a/voluptuous/schema_builder.py
+++ b/voluptuous/schema_builder.py
@@ -127,7 +127,9 @@ def default_factory(value) -> DefaultFactory:
 
 
 @contextmanager
-def raises(exc, msg: typing.Optional[str] = None, regex: typing.Optional[re.Pattern] = None) -> Generator[None, None, None]:
+def raises(
+    exc, msg: typing.Optional[str] = None, regex: typing.Optional[re.Pattern] = None
+) -> Generator[None, None, None]:
     try:
         yield
     except exc as e:
@@ -189,7 +191,9 @@ class Schema(object):
         PREVENT_EXTRA: 'PREVENT_EXTRA',
     }
 
-    def __init__(self, schema: Schemable, required: bool = False, extra: int = PREVENT_EXTRA) -> None:
+    def __init__(
+        self, schema: Schemable, required: bool = False, extra: int = PREVENT_EXTRA
+    ) -> None:
         """Create a new Schema.
 
         :param schema: Validation schema. See :module:`voluptuous` for details.
@@ -237,18 +241,17 @@ class Schema(object):
 
         Note: only very basic inference is supported.
         """
+
         def value_to_schema_type(value):
             if isinstance(value, dict):
                 if len(value) == 0:
                     return dict
-                return {k: value_to_schema_type(v)
-                        for k, v in value.items()}
+                return {k: value_to_schema_type(v) for k, v in value.items()}
             if isinstance(value, list):
                 if len(value) == 0:
                     return list
                 else:
-                    return [value_to_schema_type(v)
-                            for v in value]
+                    return [value_to_schema_type(v) for v in value]
             return type(value)
 
         return cls(value_to_schema_type(data), **kwargs)
@@ -266,8 +269,11 @@ class Schema(object):
 
     def __repr__(self):
         return "<Schema(%s, extra=%s, required=%s) object at 0x%x>" % (
-            self.schema, self._extra_to_name.get(self.extra, '??'),
-            self.required, id(self))
+            self.schema,
+            self._extra_to_name.get(self.extra, '??'),
+            self.required,
+            id(self),
+        )
 
     def __call__(self, data):
         """Validate data against this schema."""
@@ -301,24 +307,29 @@ class Schema(object):
             type_ = schema
         if type_ in (*primitive_types, object, type(None)) or callable(schema):
             return _compile_scalar(schema)
-        raise er.SchemaError('unsupported schema data type %r' %
-                             type(schema).__name__)
+        raise er.SchemaError('unsupported schema data type %r' % type(schema).__name__)
 
     def _compile_mapping(self, schema, invalid_msg=None):
         """Create validator for given mapping."""
         invalid_msg = invalid_msg or 'mapping value'
 
         # Keys that may be required
-        all_required_keys = set(key for key in schema
-                                if key is not Extra
-                                and ((self.required
-                                      and not isinstance(key, (Optional, Remove)))
-                                     or isinstance(key, Required)))
+        all_required_keys = set(
+            key
+            for key in schema
+            if key is not Extra
+            and (
+                (self.required and not isinstance(key, (Optional, Remove)))
+                or isinstance(key, Required)
+            )
+        )
 
         # Keys that may have defaults
-        all_default_keys = set(key for key in schema
-                               if isinstance(key, Required)
-                               or isinstance(key, Optional))
+        all_default_keys = set(
+            key
+            for key in schema
+            if isinstance(key, Required) or isinstance(key, Optional)
+        )
 
         _compiled_schema = {}
         for skey, svalue in schema.items():
@@ -337,7 +348,9 @@ class Schema(object):
             if type(skey) in primitive_types:
                 candidates_by_key.setdefault(skey, []).append((skey, (ckey, cvalue)))
             elif isinstance(skey, Marker) and type(skey.schema) in primitive_types:
-                candidates_by_key.setdefault(skey.schema, []).append((skey, (ckey, cvalue)))
+                candidates_by_key.setdefault(skey.schema, []).append(
+                    (skey, (ckey, cvalue))
+                )
             else:
                 # These are wildcards such as 'int', 'str', 'Remove' and others which should be applied to all keys
                 additional_candidates.append((skey, (ckey, cvalue)))
@@ -354,8 +367,10 @@ class Schema(object):
 
             # Insert default values for non-existing keys.
             for key in all_default_keys:
-                if not isinstance(key.default, Undefined) and \
-                   key.schema not in key_value_map:
+                if (
+                    not isinstance(key.default, Undefined)
+                    and key.schema not in key_value_map
+                ):
                     # A default value has been specified for this missing
                     # key, insert it.
                     key_value_map[key.schema] = key.default()
@@ -366,7 +381,9 @@ class Schema(object):
                 remove_key = False
 
                 # Optimization. Validate against the matching key first, then fallback to the rest
-                relevant_candidates = itertools.chain(candidates_by_key.get(key, []), additional_candidates)
+                relevant_candidates = itertools.chain(
+                    candidates_by_key.get(key, []), additional_candidates
+                )
 
                 # compare each given key/value against all compiled key/values
                 # schema key, (compiled key, compiled value)
@@ -431,7 +448,11 @@ class Schema(object):
 
             # for any required keys left that weren't found and don't have defaults:
             for key in required_keys:
-                msg = key.msg if hasattr(key, 'msg') and key.msg else 'required key not provided'
+                msg = (
+                    key.msg
+                    if hasattr(key, 'msg') and key.msg
+                    else 'required key not provided'
+                )
                 errors.append(er.RequiredFieldInvalid(msg, path + [key]))
             if errors:
                 raise er.MultipleInvalid(errors)
@@ -458,8 +479,7 @@ class Schema(object):
             ...   validate(Structure(one='three'))
 
         """
-        base_validate = self._compile_mapping(
-            schema, invalid_msg='object value')
+        base_validate = self._compile_mapping(schema, invalid_msg='object value')
 
         def validate_object(path, data):
             if schema.cls is not UNDEFINED and not isinstance(data, schema.cls):
@@ -547,8 +567,7 @@ class Schema(object):
          "expected str for dictionary value @ data['adict']['strfield']"]
 
         """
-        base_validate = self._compile_mapping(
-            schema, invalid_msg='dictionary value')
+        base_validate = self._compile_mapping(schema, invalid_msg='dictionary value')
 
         groups_of_exclusion = {}
         groups_of_inclusion = {}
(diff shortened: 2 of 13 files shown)


ds-cbo:
"I've left the default normalization on"

I don't have strong opinions on quotes, but I do have a strong opinion on git history: Please try to keep the changes minimal when introducing a new tool. Other developers will `git annotate` a file and 50% of the lines will get "black 2023/12/14" as annotation, which is completely useless.

voluptuous/__init__.py
```diff
@@ -1,9 +1,9 @@
 # flake8: noqa
 
+from voluptuous.error import *
 from voluptuous.schema_builder import *
-from voluptuous.validators import *
 from voluptuous.util import *
-from voluptuous.error import *
+from voluptuous.validators import *
```
ds-cbo, line 6: This might actually be a breaking change if anything in `.error` depends on anything in `.validators` (which is likely), I would disable isort for `__init__.py` files

voluptuous/schema_builder.py
```diff
@@ -148,11 +150,24 @@ def Extra(_) -> None:
...
+    bytes,
+    int,
+    str,
+    float,
+    complex,
+    type,
+    object,
+    dict,
+    None,
+    typing.Callable,
```
ds-cbo, line 170: this had intentional compact formatting (similar types on the same line), please wrap it in `# fmt: off` and `# fmt: on`

voluptuous/schema_builder.py
```diff
@@ -854,11 +869,13 @@ def is_callable(key_):
...
-                (4, is_type),  # types/classes lowest before Extra
-                (3, is_callable),  # callables after markers
-                (5, is_extra)]  # Extra lowest priority
+    priority = [
+        (1, is_remove),  # Remove highest priority after values
+        (2, is_marker),  # then other Markers
+        (4, is_type),  # types/classes lowest before Extra
+        (3, is_callable),  # callables after markers
+        (5, is_extra),
+    ]  # Extra lowest priority
```
ds-cbo, line 878: move this comment back to the respective line

voluptuous/schema_builder.py
```diff
@@ -1237,9 +1274,7 @@ def _args_to_dict(func, args):
         arg_names = func.func_code.co_varnames[:arg_count]
 
     arg_value_list = list(args)
-    arguments = dict((arg_name, arg_value_list[i])
-                     for i, arg_name in enumerate(arg_names)
-                     if i < len(arg_value_list))
+    arguments = dict((arg_name, arg_value_list[i]) for i, arg_name in enumerate(arg_names) if i < len(arg_value_list))
```
ds-cbo, line 1277: I wouldn't say that this is more readable now, please wrap it back to several lines (decreasing line length as per my other suggestion would probably help already)

voluptuous/tests/tests.py
```diff
@@ -1,26 +1,77 @@
...
+    SomeOf,
+    TooManyValid,
+    TypeInvalid,
+    Union,
+    Unordered,
+    Url,
+    UrlInvalid,
+    raises,
+    validate,
+)
```
ds-cbo, line 63: please compact this down as well, with `# fmt: off` and `# fmt: on` to keep it in place
no one likes a 50-line import statement

voluptuous/tests/tests.py
```diff
@@ -1300,39 +1319,36 @@ def __hash__(self):
...
 
 
 def test_self_all():
-    schema = Schema({"number": int,
-                     "follow": All(Self,
-                                   Schema({"extra_number": int},
-                                          extra=ALLOW_EXTRA))},
-                    extra=ALLOW_EXTRA)
+    schema = Schema(
+        {"number": int, "follow": All(Self, Schema({"extra_number": int}, extra=ALLOW_EXTRA))}, extra=ALLOW_EXTRA
```
ds-cbo, line 1343: this also didn't get any clearer, please expand it again (might be able to get black to do that with a trailing comma)

voluptuous/validators.py
```diff
@@ -1,18 +1,41 @@
...
+    MultipleInvalid,
+    NotEnoughValid,
+    NotInInvalid,
+    PathInvalid,
+    RangeInvalid,
+    TooManyValid,
+    TrueInvalid,
+    TypeInvalid,
+    UrlInvalid,
+)
```
ds-cbo, line 35: please compact this one down as well, 20-line imports are still no fun

ds-cbo (changes requested):

voluptuous/schema_builder.py
alecthomas: Agreed.

alecthomas:
> Counterpoint: I often have 2 terminals side-to-side on my 16:9 screen, and then it only fits up to 100 characters comfortably. It also depends on font-size, which is still very personal nowadays. I would keep the default value of black (80 + 10% wrap margin)

Counter-counterpoint: I suspect the majority of people, myself included, prefer 100 or 120 columns, and for the minority who don't, having to scroll horizontally every now and then is not the end of the world. I have to do it myself sometimes on smaller screens, and I'm fine with it.

voluptuous/schema_builder.py
```diff
@@ -937,8 +954,7 @@ class Msg(object):
 
     def __init__(self, schema: Schemable, msg: str, cls: typing.Optional[typing.Type[Error]] = None) -> None:
         if cls and not issubclass(cls, er.Invalid):
-            raise er.SchemaError("Msg can only use subclases of"
-                                 " Invalid as custom class")
+            raise er.SchemaError("Msg can only use subclases of" " Invalid as custom class")
```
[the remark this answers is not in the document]
ds-cbo: I've had this before, the thing with black is that it 100% guarantees that the AST stays the exact same between transformations, and merging two strings would break that promise on python 3.7- (see https://github.com/psf/black/issues/26 )

But I think it should detect that we're not running on python 3.7- and merge them, so I'm not sure what's going on there

voluptuous/schema_builder.py
```diff
@@ -296,24 +313,22 @@ def _compile(self, schema):
...
-        all_required_keys = set(key for key in schema
-                                if key is not Extra
-                                and ((self.required
-                                      and not isinstance(key, (Optional, Remove)))
-                                     or isinstance(key, Required)))
+        all_required_keys = set(
+            key
+            for key in schema
+            if key is not Extra
+            and ((self.required and not isinstance(key, (Optional, Remove))) or isinstance(key, Required))
```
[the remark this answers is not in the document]
antoni-szych-rtbhouse: I'm afraid it's not configurable. Black has (by design and purpose) very limited configuration options, boiling down to:
- line length
- string quotes normalization on or off
- "magic trailing comma" on or off

The upside is that you don't have to choose and maintain your own formatting style in a project.

antoni-szych-rtbhouse:
> "I've left the default normalization on"
> 
> I don't have strong opinions on quotes, but I do have a strong opinion on git history: Please try to keep the changes minimal when introducing a new tool. Other developers will `git annotate` a file and 50% of the lines will get "black 2023/12/14" as annotation, which is completely useless.

This can be actually [easily fixed both locally and for github blame](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame): we just need to put the commit sha into the `.git-blame-ignore-revs` file. Note that it will have to be added after merging this PR, as we don't know the merge commit sha in advance :)

voluptuous/validators.py
```diff
@@ -33,42 +56,39 @@
...
-    re.IGNORECASE
+    re.IGNORECASE,
 )
 DOMAIN_REGEX = re.compile(
     # start anchor, because fullmatch is not available in python 2.7
     "(?:"
     # domain
-    r'(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)+'
-    r'(?:[A-Z]{2,6}\.?|[A-Z0-9-]{2,}\.?$)'
+    r"(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)+" r"(?:[A-Z]{2,6}\.?|[A-Z0-9-]{2,}\.?$)"
```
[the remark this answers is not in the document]
antoni-szych-rtbhouse: sure. Note that this regex is long obsolete due to tlds like `.basketball`, `.google` or `.youtube`:
- https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains#ICANN-era_generic_top-level_domains
- https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains#Brand_top-level_domains

voluptuous/schema_builder.py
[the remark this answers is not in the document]
antoni-szych-rtbhouse: You're right, it does not out of the box [(yet, but it should starting with black 24)](https://github.com/psf/black/issues/1837). I will run it with the mentioned experimental flag and manually check output before commiting.

[flake8-implicit-str-concat](https://pypi.org/project/flake8-implicit-str-concat/) would be useful for such cases, I can add it to the flake8 tox env if you agree.
I'm using pylint on a daily basis, it also [has this rule](https://pylint.readthedocs.io/en/stable/user_guide/messages/warning/implicit-str-concat.html)

voluptuous/validators.py
antoni-szych-rtbhouse: yep, there is, `multi_line_output = 5` - however it's incompatible with black if it gets longer than `line_length`. `# fmt: off` and `# fmt: on` to exclude it from black formatting gets the job done.

voluptuous/tests/tests.py
antoni-szych-rtbhouse: sure thing

voluptuous/schema_builder.py
antoni-szych-rtbhouse: limiting the line length to default 88 did the trick

voluptuous/tests/tests.py
antoni-szych-rtbhouse: limiting the line length to default 88 did the trick

voluptuous/schema_builder.py
antoni-szych-rtbhouse: sure thing.

voluptuous/schema_builder.py
antoni-szych-rtbhouse: sure thing

voluptuous/__init__.py
antoni-szych-rtbhouse:
```
from voluptuous.error import * # isort: skip
```
made `error` go to the bottom of imports

antoni-szych-rtbhouse:
Hi @ds-cbo @alecthomas can you please take a look at the updated version? :)

ds-cbo:
I have no new complaints, but I'm also not responsible for this project

antoni-szych-rtbhouse:
@alecthomas could you please take a look? :) I think all the major obstacles were resolved with the change in line length and few `fmt:off/on`s and `ignore`s.

alecthomas:
Mmm yeah, one last change please: can you make the default string quoting `'` instead of `"`. That is the default in this project, and leaving it as-is will reduce the diff size considerably.

voluptuous/validators.py
```diff
@@ -33,42 +40,42 @@
     # start anchor, because fullmatch is not available in python 2.7
     "(?:"
     # dot-atom
-    r"(^[-!#$%&'*+/=?^_`{}|~0-9A-Z]+"
-    r"(\.[-!#$%&'*+/=?^_`{}|~0-9A-Z]+)*$"
+    r"(^[-!#$%&'*+/=?^_`{}|~0-9A-Z]+(\.[-!#$%&'*+/=?^_`{}|~0-9A-Z]+)*$"
```
alecthomas, line 43: Why did this get joined? Can we split this again please?

voluptuous/validators.py
```diff
@@ -33,42 +40,42 @@
...
     # start anchor, because fullmatch is not available in python 2.7
     "(?:"
     # domain
     r'(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)+'
+    # tld
     r'(?:[A-Z]{2,6}\.?|[A-Z0-9-]{2,}\.?$)'
     # literal form, ipv4 address (SMTP 4.1.3)
-    r'|^\[(25[0-5]|2[0-4]\d|[0-1]?\d?\d)'
-    r'(\.(25[0-5]|2[0-4]\d|[0-1]?\d?\d)){3}\]$'
+    r'|^\[(25[0-5]|2[0-4]\d|[0-1]?\d?\d)(\.(25[0-5]|2[0-4]\d|[0-1]?\d?\d)){3}\]$'
```
alecthomas, line 59: Same thing here.

alecthomas (changes requested):

voluptuous/validators.py
antoni-szych-rtbhouse: because these 2 lines can fit into one line, black merged them into:
```
    r"(^[-!#$%&'*+/=?^_`{}|~0-9A-Z]+" r"(\.[-!#$%&'*+/=?^_`{}|~0-9A-Z]+)*$"
```
which raises obvious issue "why the strings were not concatenated, which results in:
```
    r"(^[-!#$%&'*+/=?^_`{}|~0-9A-Z]+(\.[-!#$%&'*+/=?^_`{}|~0-9A-Z]+)*$"
```
The only way to keep the original version is `# fmt: on` and `# fmt: off`. I've added it, so these lines are back to being separate.

voluptuous/validators.py
antoni-szych-rtbhouse: I've added `# fmt: on` and `# fmt: off`, so these lines are back to being separate.

antoni-szych-rtbhouse:
> Mmm yeah, one last change please: can you make the default string quoting `'` instead of `"`. That is the default in this project, and leaving it as-is will reduce the diff size considerably.

Sure, I've removed the string normalization commit (it was the last one) and the PR is smaller (`+755 −433` vs `+1,047 −730` previously).

I've also made a test and replaced all possible double quotes to single quotes in 231 lines. This would cause the diff to be `+966 −648`. Please let me know if case you would like me to commit this.

alecthomas (approved):
Thanks for doing this!

ds-cbo:
> This can be actually [easily fixed both locally and for github blame](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame): we just need to put the commit sha into the .git-blame-ignore-revs file. Note that it will have to be added after merging this PR, as we don't know the merge commit sha in advance :)

Now please don't forget to do that as well, now we have this:

image

antoni-szych-rtbhouse:
> Now please don't forget to do that as well, now we have this:

Done in #505, it will look like following (screenshot from [my fork with this commit already merged](https://github.com/antoni-szych-rtbhouse/voluptuous/blame/master/voluptuous/humanize.py)):

image


As you can see, there are some lines annotated with the refactor commit, however these are only the new lines, which did not exist at all previously. However the important lines like 26-31, 36-40, 45-46, 51-53 show the pre-refactor annotation.

PR #512, opened 2024-06-06
Add Any type to defaults

elupus:
These function ended up as only accepting UNDEFINED as argument since the function was partially typed.


farmio (commented):
same for `Required`

elupus:
Fixed

alecthomas (approved):
Awesome thanks.

PR #516, opened 2024-06-25
Improve validator typing to allow non-number formats for min and max

cdce8p:
`NullableNumber` is defined as `int | float | None`. This prevents the use of other values for `min / max` in `Range`, `Clamp` and `Length` which also support rich comparisons. E.g. `timedelta`.

Use `SupportsAllComparisons` from `_typeshed` instead.

/CC @bdraco


bdraco (approved):

bdraco (approved):
Nice fix

alecthomas (approved):

PR #519, opened 2024-06-25
Don't enforce type for unused description attribute

cdce8p:
> `description` is an optional field, unused by Voluptuous itself, but can be
    introspected by any external tool, for example to generate schema documentation.

As per the docstring, the `description` attribute is unused by voluptuous itself but can be used by external tools. It therefore shouldn't force the type to be `str | None`. If the tool prefers `dict`, that should be valid as well.

### DIFF
--- a/voluptuous/schema_builder.py
+++ b/voluptuous/schema_builder.py
@@ -1032,7 +1032,7 @@ class Marker(object):
         self,
         schema_: Schemable,
         msg: typing.Optional[str] = None,
-        description: typing.Optional[str] = None,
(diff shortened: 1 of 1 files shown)


bdraco (approved):

cdce8p:
@alecthomas If you have some time, it would be great if you could take a look at this one and #521. These two are the last remaining changes needed for Home Assistant.

After they are merged, a new release (`0.15.1`) would be greatly appreciated.

bdraco:
Thank you !!

PR #520, opened 2024-06-26
Use TypeVar for Schema.schema type

cdce8p:
`Schemable` is great as a catch all case. However, the Union prevents further uses especially of the `schema` attribute.
E.g.
```py
schema = vol.Schema({"key": str})
schema.schema.items()  # should access the dict items used to create the schema
```

This PR changes the `schema` attribute type to a `TypeVar`. This allows for either implicit or explicit specialization by the user. It would fallback to `typing.Any` if type checkers can't infer the specialized type.

_That's still better than the current situation as that just always emits an error._
```py
error: Item "type" of "Schema | Object | Mapping[Any, Any] | list[Any] | tuple[Any, ...] |
    <12 more items> | None" has no attribute "__iter__" (not iterable)  [union-attr]
```

### DIFF
--- a/voluptuous/schema_builder.py
+++ b/voluptuous/schema_builder.py
@@ -162,8 +162,10 @@ Schemable = typing.Union[
 ]
 # fmt: on
 
+_SchemableT = typing.TypeVar("_SchemableT", bound=Schemable)
 
-class Schema(object):
+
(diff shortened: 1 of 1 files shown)


cdce8p:
Closed in favor of #521

PR #522, opened 2024-06-28
Fix regression with ALLOW_EXTRA and Any validator

cdce8p:
Another regression from #479 found while testing the `0.15.1` in https://github.com/home-assistant/core/pull/120631.

@alecthomas Would be great to have a `0.15.2` with that 🙏🏻

/CC @bdraco


bdraco (approved):
Makes sense that it needed to be one deeper in the block


Thanks @cdce8p

PR #534, opened 2025-09-01
Feature: Support requiring anyOf a list of keys

cibernox:
This adds a new feature to Voluptuous, which is somewhat akin to what json-schema does with the special key `anyOf`.

    `Schema({Required(Any('color', 'temperature', 'brightness')): str})` will validate that AT LEAST ONE of these three values is present. That doesn't preclude any individual validation on each of those fields to still apply.
    That means that in the above example, if `color` is present, brightness doesn't need to be present. But if brightness is present, all other validations of brightness (like checking that its value is a number between 0 and 100) still apply.

### DIFF
--- a/voluptuous/tests/tests.py
+++ b/voluptuous/tests/tests.py
@@ -2190,3 +2190,39 @@ def test_required_complex_key_value_validation():
     assert "expected str" in error_msg
 
 
+def test_complex_required_keys_with_specific_value_validation():
+    """Test complex required keys combined with specific value validation for brightness range."""
+    schema = Schema({
+        Required(Any('color', 'temperature', 'brightness')): str,
+        'brightness': All(Coerce(int), Range(min=0, max=100)),  # Additional validation for brightness specifically
+        'device_id': str
+    })
+    
+    # Valid - color provided, no brightness validation needed
+    result = schema({'color': 'red', 'device_id': 'light1'})
+    assert result == {'color': 'red', 'device_id': 'light1'}
+    
+    # Valid - temperature provided, no brightness validation needed
+    result = schema({'temperature': '3000K', 'device_id': 'light1'})
+    assert result == {'temperature': '3000K', 'device_id': 'light1'}
+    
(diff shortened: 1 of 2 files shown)


cibernox:
Is this something that you're interested in? From the readme, i'm not sure how amenable you'd be to this. I'm not sure if this qualifies as a new feature, as it's using roughly the existing building blocks to wrap in `vol.Required` complex keys like `vol.Any`.

None of the pieces is new, but it enables a new kind of validation that wasn't possible before.

alecthomas:
Hey! Yep, sounds reasonable to me, thanks

voluptuous/tests/tests.py
```diff
@@ -2037,3 +2037,282 @@ def test_humanize_error_with_none_data():
 
     error_message = humanize_error(data, ctx.value)
     assert "expected a dictionary" in error_message
+
```
alecthomas: This is a lot of tests, most of which seem redundant. Can you reduce the number of tests to the minial set please?

cibernox: sure thing!

cibernox: Done! I reduced que number of tests and eliminated some asserts that were kind of redundant with other ones.

There's still a fair amount of assertions but because I wanted to be thorough in that we weren't regressing in any subtle way.

cibernox:
Thank you sir 🙇 

May we get a new release cut?

cibernox:
@alecthomas ping on pushing this new version to pypi 👍

cibernox:
FWIW, I brought this up with HomeAssistant / Open Home Foundation and they are going to discuss if they want / can become maintainers.

alecthomas:
Just to be clear - I'd add them as maintainers to the GitHub project and PyPi project

cibernox:
Noted. I was just the messenger.

spacegaier:
> @alecthomas ping on pushing this new version to pypi 👍

I will prepare a new 0.16.0 release

PR #537, opened 2026-05-28
Feature: Support enum members as scalar schemas and mapping keys

epenet:
Allow an enum.Enum member to be used directly as a schema (matched by equality) and as a mapping key, by compiling it as a scalar.

Spotted in https://github.com/home-assistant/core/pull/172425 and https://github.com/home-assistant/architecture/discussions/1406


alecthomas (approved):

alecthomas:
Thanks!

PR #538, opened 2026-06-21
Allow Schema.extend to accept Schema instances

### LINKED ISSUE #386: Define property with list of references
Hi! I am interested in, is it possible to define schemas in voluptuous like in openapi 3.0.0
Here is an example of openapi:

```
data:
  allOf:
    - $ref: '#/components/schemas/Project'
    - $ref: '#/components/schemas/WithProvider'
    - $ref: '#/components/schemas/WithGallery'
```

I talk about something like:

```
Schema(
  {
    'data': All({
      Project,
      WithProvider,
      WithGallery
    })
  }
)
```

may be not like this, maybe like that:

```
Schema({
    'data': {
      $ref: Project,
      $ref: WithProvider,
      $ref: WithGallery
    }
  }
)
```

or maybe like this:
```
Schema({
    'data': {
      Project,
      WithProvider,
      [WithGallery]
    }
  }
)
```

I think that this feature with list of references inside of property could be very helpful for the whole community to define more schemas!

Thank you!
Best regards, Aleksandr

aleksandr-kotlyar:
## Summary

This PR allows `Schema.extend()` to accept another dictionary-based `Schema` instance.

Previously, `Schema.extend()` only accepted raw dictionary fragments:

```python
base = Schema({"a": int})
extended = base.extend({"b": str})
```

Passing a `Schema` instance failed because `extend()` asserted that the argument itself was a dictionary:

```python
base = Schema({"a": int})
child = Schema({"b": str})

base.extend(child)
# AssertionError: Both schemas must be dictionary-based
```

This PR makes the following work:

```python
base = Schema({"a": int})
child = Schema({"b": str})

extended = base.extend(child)
assert extended.schema == {"a": int, "b": str}
```

## Required semantics

When the extension is a `Schema`, this PR preserves the child schema’s `required` behavior by normalizing plain keys before merging.

If the child schema is required, plain child keys become explicitly required:

```python
base = Schema({"a": int})
child = Schema({"b": str}, required=True)

extended = base.extend(child)
extended({"a": 1})
# raises MultipleInvalid: required key not provided @ data['b']
```

If the child schema is not required, plain child keys remain optional even when the base schema is required:

```python
base = Schema({"a": int}, required=True)
child = Schema({"b": str}, required=False)

extended = base.extend(child)
assert extended({"a": 1}) == {"a": 1}
```

The same normalization is applied recursively to nested dictionary values.

## Preserved behavior

This PR keeps existing raw dict behavior unchanged:

```python
base = Schema({"a": int}, required=True)
extended = base.extend({"b": str})
```

As before, the raw dict extension is merged directly and interpreted using the resulting schema’s settings.

Explicit marker keys are preserved and are not rewritten:

```python
Required("a")
Optional("b")
Remove("c")
Extra
```

The return type is still `type(self)`, so subclasses of `Schema` continue to work.

## Extra handling

This PR does not try to merge `extra` semantics from the extension schema.

If the extension `Schema.extra` differs from the resulting schema’s `extra`, `Schema.extend()` raises `SchemaError` with guidance to pass `child_schema.schema` explicitly when raw dict merge semantics are desired.

## Tests

Added regression coverage for:

- extending with a dictionary-based `Schema`;
- child `Schema(..., required=True)` preserving required keys;
- child `Schema(..., required=False)` staying optional under a required base schema;
- explicit `Required`, `Optional`, `Remove`, and `Extra` marker preservation;
- recursive nested dict required/optional normalization;
- incompatible `extra` settings raising `SchemaError`;
- non-dictionary `Schema` instances still raising the existing assertion;
- existing raw dict extension behavior;
- existing subclass return behavior.

Focused test command:

```bash
./.venv/bin/pytest voluptuous/tests/tests.py -k "schema_extend or subschema_extension"
```

Result:

```text
14 passed, 162 deselected
```

Closes #386

### DIFF
--- a/README.md
+++ b/README.md
@@ -539,6 +539,25 @@ requirements. In that case you can use `Schema.extend` to create a new
 ['age', 'name']
 ```
 
+`Schema.extend` can also accept another dictionary-based `Schema`. In that case
+the extension schema keeps its own `required` setting:
+
+```pycon
+>>> person = Schema({'name': str}, required=True)
+>>> contact = Schema({'email': str})
+>>> person_with_contact = person.extend(contact)
+>>> person_with_contact({'name': 'Ada'})
+{'name': 'Ada'}
+>>> person_with_contact({'name': 'Ada', 'email': 'ada@example.com'})
+{'name': 'Ada', 'email': 'ada@example.com'}
+>>> required_contact = Schema({'phone': str}, required=True)
+>>> person_with_phone = person.extend(required_contact)
+>>> person_with_phone({'name': 'Ada'})
+Traceback (most recent call last):
+...
+voluptuous.error.MultipleInvalid: required key not provided @ data['phone']
+```
+
 The original `Schema` remains unchanged.
 
 ### Objects
--- a/voluptuous/schema_builder.py
+++ b/voluptuous/schema_builder.py
@@ -716,10 +716,26 @@ class Schema(object):
 
         Both schemas must be dictionary-based.
 
-        :param schema: dictionary to extend this `Schema` with
+        :param schema: dictionary, or Schema wrapping a dictionary, to extend
+            this `Schema` with
         :param required: if set, overrides `required` of this `Schema`
         :param extra: if set, overrides `extra` of this `Schema`
         """
+
+        result_required = required if required is not None else self.required
+        result_extra = extra if extra is not None else self.extra
+
+        if isinstance(schema, Schema):
+            if schema.extra != result_extra:
+                raise er.SchemaError(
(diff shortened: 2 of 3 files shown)


aleksandr-kotlyar:
Hi @spacegaier, sorry for the ping.

This PR is ready for review when you have time. It keeps the change focused on `Schema.extend()` accepting dictionary-based `Schema` instances, includes regression tests, and documents the new behavior.

I would appreciate any feedback on whether this approach matches the project’s expected semantics, especially around `required` and `extra` handling.

Thanks!

alecthomas (approved):
Makes sense, I like it.

PR #543, opened 2026-07-03
Raise Invalid for non-finite Number values

### LINKED ISSUE #541: Number validator leaks a raw TypeError on Infinity/NaN instead of raising Invalid
### Summary
The `Number` validator (with `precision`/`scale`) converts its input to `Decimal` and inspects precision/scale. For `Decimal('Infinity')` / `Decimal('NaN')`, that inspection raises a raw `TypeError("infinity and NaN have no precision")`, which escapes the validator instead of being surfaced as a voluptuous `Invalid`. Callers that expect validators to raise only `Invalid` get an unexpected `TypeError`.

### Reproduction
```python
from voluptuous import Number
Number(precision=6, scale=2)("Infinity")
# TypeError: infinity and NaN have no precision
```

### Expected
A non-finite numeric input is rejected with `Invalid` (the validator contract), not a raw `TypeError`.

### Actual
An uncaught `TypeError` leaks out of the validator.

### Fix sketch
Detect the non-finite case and `raise Invalid(...)`.

### Environment
voluptuous 0.16.0 (master @ 87825d6), Python 3.12.

CodingFeng101:
Fixes #541.

## What changed

`Number` now rejects non-finite Decimal values such as `Infinity` and `NaN` with `Invalid` instead of leaking the internal `TypeError` from precision inspection.

## Why

`Number` is a validator and callers expect validation failures to surface as voluptuous `Invalid` / `MultipleInvalid`. Non-finite numeric strings currently escape that contract when precision or scale checking is enabled.

## Validation

- `python -m pytest voluptuous\tests\tests.py::test_number_validation_with_string voluptuous\tests\tests.py::test_number_validation_with_non_finite_number voluptuous\tests\tests.py::test_number_validation_with_invalid_precision_invalid_scale -q`
- `python -m pytest voluptuous\tests\tests.py -q`
- `python -m py_compile voluptuous\validators.py voluptuous\tests\tests.py`
- `git diff --check`

### DIFF
--- a/voluptuous/tests/tests.py
+++ b/voluptuous/tests/tests.py
@@ -1010,6 +1010,16 @@ def test_number_validation_with_invalid_precision_invalid_scale():
         assert False, "Did not raise Invalid for String"
 
 
+def test_number_validation_with_non_finite_number():
+    schema = Schema({"number": Number(precision=6, scale=2)})
+    try:
+        schema({"number": 'Infinity'})
+    except MultipleInvalid as e:
+        assert str(e) == "Value has no precision for dictionary value @ data['number']"
+    else:
+        assert False, "Did not raise Invalid for Infinity"
+
+
 def test_number_validation_with_valid_precision_scale_yield_decimal_true():
     """Test with Number with valid precision and scale"""
     schema = Schema({"number": Number(precision=6, scale=2, yield_decimal=True)})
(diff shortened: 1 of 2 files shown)


alecthomas:
Thanks, but #542 already addresses this and was raised first.

PR #545, opened 2026-07-04
Emit missing-required-key errors in schema-declaration order (#484)

### LINKED ISSUE #484: MultipleInvalid errors order is not always the same
When validating this schema:

```python
Schema1 = Schema({
    Required('value1'): Match2(r'^[\d\w_-]{1,64}$'),
    Required('value2'): Any(IntField(), DecimalField(precision=3), TemplateString(), str)
})
``` 

With this test:

```python
def test_schema1_empty():
    with pytest.raises(MultipleInvalid) as exception_info:
        Schema1({})
    assert len(exception_info.value.errors) == 2
    assert str(exception_info.value.errors[0]) == "required key not provided @ data['value1']"
    assert str(exception_info.value.errors[1]) == "required key not provided @ data['value2']"
``` 

The test fails 1 over 3-4 times.
I found it is related to the use of a set, that is unordered in python in schema_builder.py

Current code:
```python
    def _compile_mapping(self, schema, invalid_msg=None):
        """Create validator for given mapping."""
        invalid_msg = invalid_msg or 'mapping value'

        # Keys that may be required
        all_required_keys = set(key for key in schema
                                if key is not Extra
                                and ((self.required
                                      and not isinstance(key, (Optional, Remove)))
                                     or isinstance(key, Required)))

        # Keys that may have defaults
        all_default_keys = set(key for key in schema
                               if isinstance(key, Required)
                               or isinstance(key, Optional))
``` 

My proposal (that works for me) would be to use a dict instead of a set (to preserve validations errors order):
```python
    def _compile_mapping(self, schema, invalid_msg=None):
        """Create validator for given mapping."""
        invalid_msg = invalid_msg or 'mapping value'

        # Keys that may be required

        all_required_keys = {}
        for key in schema:
            if key is not Extra \
                and ((self.required \
                and not isinstance(key, (Optional, Remove))) \
                or isinstance(key, Required)):
                    all_required_keys[key] = key

        # Keys that may have defaults
        all_default_keys = {}
        for key in schema:
            if isinstance(key, Required) or isinstance(key, Optional):
                all_default_keys[key] = key
```

uttam12331:
## Summary

Closes #484.

Required keys were collected into a `set`, so the `MultipleInvalid` raised for missing required keys came out in an arbitrary, hash-seed-dependent order. The reporter's two-key test failed ~1 in 3 runs:

```python
S = Schema({Required('value1'): str, Required('value2'): str, Required('value3'): str})
S({})  # errors sometimes [value1, value2, value3], sometimes [value2, value1, value3], ...
```

## Change

Store the required keys in an insertion-ordered `dict` (used as an ordered set) instead of a `set`, and replace the `set.discard(...)` calls with `dict.pop(..., None)`. The leftover-required-key errors are now emitted in the order the keys were declared in the schema. No other behavior changes.

Verified across several `PYTHONHASHSEED` values — before, the error order varied per seed; after, it is always declaration order.

## Tests

Added `test_required_key_error_order_is_stable` (5 required keys, asserts the error paths equal the declaration order — an accidental correct ordering on the old code would be ~1 in 120). Full suite passes (170 passed); `black --check` and `flake8` clean. Added a CHANGELOG entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **Fixes**:
 
 * [#539](https://github.com/alecthomas/voluptuous/pull/539): Raise `Invalid` instead of leaking `TypeError`/`ValueError` for non-numeric input to the `Number` validator
+* [#484](https://github.com/alecthomas/voluptuous/issues/484): Emit `required key not provided` errors in a stable schema-declaration order
 
 ## [0.16.0]
 
--- a/voluptuous/schema_builder.py
+++ b/voluptuous/schema_builder.py
@@ -240,23 +240,25 @@ class Schema(object):
         """Create validator for given mapping."""
         invalid_msg = invalid_msg or 'mapping value'
 
-        # Keys that may be required
-        all_required_keys = set(
-            key
+        # Keys that may be required. A dict is used as an insertion-ordered set
+        # so that "required key not provided" errors are emitted in a stable,
+        # schema-declaration order rather than an arbitrary set order (#484).
+        all_required_keys = {
+            key: None
             for key in schema
             if key is not Extra
             and (
                 (self.required and not isinstance(key, (Optional, Remove)))
                 or isinstance(key, Required)
             )
-        )
+        }
 
         # Complex required keys that need special validation
-        complex_required_keys = set(
-            key
+        complex_required_keys = {
+            key: None
             for key in all_required_keys
             if isinstance(key, Required) and key.is_complex_key
-        )
+        }
 
         # Keys that may have defaults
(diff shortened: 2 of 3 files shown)


alecthomas (approved):

alecthomas:
This has conflicts that will need to be resolved.

uttam12331:
Thanks @alecthomas — rebased on latest `master` and resolved the CHANGELOG conflict (moved the #484 entry under the new `[Unreleased]` section). Full suite still passes (182 passed). Ready for merge.

alecthomas:
@codex review

PR #550, opened 2026-08-02
Accept Mapping inputs for dict schemas

### LINKED ISSUE #299: Validating a Mapping that isn't a dict fails
This is essentially the same issue as #125; it appears maybe this feature is now broken?

I, like #125, am attempting to validate an aiohttp request query arguments, which isn't a `dict`, but is a `Mapping`.

However, `voluptuous` requires a `dict`, as can be seen [here](https://github.com/alecthomas/voluptuous/blob/77ee5da9e33ccd0ec68f2b670d8ef0569c89041f/voluptuous/schema_builder.py#L499-L500); I see this in my stack trace:

```
  File "…/lib/python3.6/site-packages/voluptuous/schema_builder.py", line 221, in __call__
    return self._compiled([], data)
  File "…/lib/python3.6/site-packages/voluptuous/schema_builder.py", line 504, in validate_dict
    raise er.DictInvalid('expected a dictionary', path)
voluptuous.error.DictInvalid: expected a dictionary

During handling of the above exception, another exception occurred:

  File "…/lib/python3.6/site-packages/aiohttp/web_protocol.py", line 422, in start
    resp = yield from self._request_handler(request)
  File "…/lib/python3.6/site-packages/aiohttp/web.py", line 306, in _handle
    resp = yield from handler(request)
  File "…/__main__.py", line 32, in wrapper
    json_data = await func(*args, **kwargs)
  File "…/__main__.py", line 88, in foo
    validated_qargs = _SCHEMA(request.query)
  File "…/lib/python3.6/site-packages/voluptuous/schema_builder.py", line 225, in __call__
    raise er.MultipleInvalid([e])
voluptuous.error.MultipleInvalid: expected a dictionary
```

I think it's as simple as changing that `dict` to `Mapping`?

deepakganesh78:
Fixes #299

## Reproduction
A dictionary schema currently rejects `collections.abc.Mapping` implementations that are not `dict` subclasses:

```python
from collections.abc import Mapping
from voluptuous import Schema

class QueryArgs(Mapping):
    def __init__(self, data):
        self._data = data
    def __getitem__(self, key):
        return self._data[key]
    def __iter__(self):
        return iter(self._data)
    def __len__(self):
        return len(self._data)

Schema({"page": int})(QueryArgs({"page": 1}))
```

On current `master` this raises `MultipleInvalid: expected a dictionary`.

## Root cause
`Schema._compile_dict()` only accepted `dict` instances even though the schema type already treats `collections.abc.Mapping` as a valid dictionary-like schema and mappings expose the `.items()` / membership APIs used by validation.

## Fix
Accept `collections.abc.Mapping` inputs in dictionary schemas. Existing `dict` and `dict` subclass output behavior is preserved; generic `Mapping` inputs produce a plain `dict`, avoiding assumptions that arbitrary mapping classes can be constructed empty and mutated.

## Compatibility notes
The existing error message for non-mapping inputs remains `expected a dictionary`. This only broadens accepted inputs for objects implementing the standard Mapping protocol.

## Validation
- Minimal reproducer failed before the fix with `MultipleInvalid: expected a dictionary` and succeeds after the fix.
- Regression check with fix reverted: `python -m pytest voluptuous\tests\tests.py::test_schema_accepts_mapping -q` fails with `MultipleInvalid: expected a dictionary`.
- Targeted regression with fix: `python -m pytest voluptuous\tests\tests.py::test_schema_accepts_mapping -q` -> 1 passed.
- Full suite: `python -m pytest` -> 183 passed.
- Formatting/lint/type checks: `black --check voluptuous\schema_builder.py voluptuous\tests\tests.py` -> unchanged; `isort --check voluptuous\schema_builder.py voluptuous\tests\tests.py` -> passed; `flake8 --doctests voluptuous\schema_builder.py voluptuous\tests\tests.py` -> passed; `mypy voluptuous` -> success, no issues found in 10 source files (mypy 2.3.0 also notes the repo's configured `python_version = 3.9` is below its supported range on Python 3.13).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * [#539](https://github.com/alecthomas/voluptuous/pull/539): Raise `Invalid` instead of leaking `TypeError`/`ValueError` for non-numeric input to the `Number` validator
 * [#542](https://github.com/alecthomas/voluptuous/pull/542): Raise `Invalid` instead of leaking a raw `TypeError` for `Infinity`/`NaN` input to the `Number` validator
+* [#299](https://github.com/alecthomas/voluptuous/issues/299): Accept `Mapping` instances, not only `dict`, when validating dictionary schemas
 
 ## [0.16.0]
 
--- a/voluptuous/schema_builder.py
+++ b/voluptuous/schema_builder.py
@@ -531,7 +531,7 @@ class Schema(object):
                 g.append(node)
 
         def validate_dict(path, data):
-            if not isinstance(data, dict):
+            if not isinstance(data, collections.abc.Mapping):
                 raise er.DictInvalid('expected a dictionary', path)
 
             errors = []
@@ -572,7 +572,7 @@ class Schema(object):
             if errors:
                 raise er.MultipleInvalid(errors)
 
-            out = data.__class__()
+            out = data.__class__() if isinstance(data, dict) else {}
             return base_validate(path, data.items(), out)
 
         return validate_dict
--- a/voluptuous/tests/tests.py
+++ b/voluptuous/tests/tests.py
@@ -991,6 +991,25 @@ def test_schema_empty_dict():
         assert False, "Did not raise correct Invalid"
 
 
+def test_schema_accepts_mapping():
+    class QueryArgs(collections.abc.Mapping):
+        def __init__(self, data):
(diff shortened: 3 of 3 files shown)


alecthomas:
@codex review

PR #552, opened 2026-08-17
Fix: In _compile_sequence's validate_sequence (schema_builder.py), the loop...

M001N:
## Summary
Instead of trying to reverse-engineer 'is this a key-shape-mismatch error' from an error's class/message downstream, validate_mapping now tags every error it constructs that genuinely represents 'this candidate key didn't match the data's key at all' with a `_key_shape_mismatch = True` marker attribute, at each of the three points such errors are created: (1) when a key candidate's ckey() call rejects the key (covers key validators like In/Match generically, not just literal keys), (2) when no candidate matched and 'extra keys not allowed' is raised, (3) when a required key is missing entirely. _is_key_shape_mismatch in schema_builder.py now just checks for that marker plus depth, so it generically covers every source of 'key didn't match' rather than special-casing specific error text/classes. Value-validation errors (a key that WAS matched but whose value is wrong) are never tagged, so the heuristic still correctly aborts the alternative search for genuine value-type errors like Case 1 in the issue.

## Problem
alecthomas/voluptuous issue reference: alecthomas/voluptuous#142

## Root Cause
In _compile_sequence's validate_sequence (schema_builder.py), the loop over schema alternatives for a list item aborts early via `if len(e.path) > len(index_path): raise` whenever an alternative's error path is one level deeper than the item's own path -- intended to mean 'top-level shape matched, nested detail is wrong, surface it immediately'. But validate_mapping (the dict validator) always reports key-related failures ('extra keys not allowed', missing required key, or a key validator/schema like In()/Match() rejecting the key) exactly one level deeper than the path it was given, even when the alternative's shape doesn't match the data at all. The prior attempt patched this by recognizing two specific error shapes (plain 'extra keys not allowed' Invalid and RequiredFieldInvalid) via message/class inspection in a new _is_key_shape_mismatch helper, but this missed the case where a dict alternative uses a key validator (In, Match, etc.) instead of a literal key: that rejection surfaces as the key validator's own differently-classed, differently-worded Invalid (e.g. InInvalid) at the same depth, which the message/class-based check couldn't recognize as a shape mismatch.

## Testing
PASS - full suite: 183 passed (including tests.md doctests); test_fix_142_list_of_dict_alternatives extended to cover: Case 0 (issue's reported bug) success, Case 1 (genuine value-type error) still correctly raised, 3-alternative case, Required-key shape mismatch, and the new key-validator cases (In and Match, both success and a real value-error-not-swallowed case). The 2 pre-existing --doctest-modules failures (Schema._compile_dict, Remove) were confirmed identical on both the base commit and this change via git stash comparison, so they are unrelated pre-existing issues, not a regression.

## Related Issue
alecthomas/voluptuous#142

### DIFF
--- a/voluptuous/schema_builder.py
+++ b/voluptuous/schema_builder.py
@@ -343,6 +343,14 @@ class Schema(object):
                     except er.Invalid as e:
                         if len(e.path) > len(key_path):
                             raise
+                        # This candidate key schema/validator rejected the
+                        # provided key outright -- the data's key has
+                        # nothing to do with this candidate, as opposed to
+                        # a key that matched but whose value was wrong.
+                        # Tag it so callers such as _compile_sequence's
+                        # list-of-dict-alternatives heuristic can tell the
+                        # two situations apart.
+                        e._key_shape_mismatch = True
                         if not error or len(e.path) > len(error.path):
                             error = e
                         continue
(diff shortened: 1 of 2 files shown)


alecthomas:
@codex review

